PR #5, opened 2016-07-18
logging: switch to glog

toanju:
* all logging statements are updated
* spec file updated


src/baseboxd/baseboxd.cpp
```diff
@@ -3,12 +3,11 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 #include "cunixenv.hpp"
-#include "roflibs/netlink/clogging.hpp"
+#include <glog/logging.h>
```
octopusx: This should be re-ordered, since it's not a local source reference anymore. 
But if we want to re-order the includes in basebox code in general, this might be a topic for an entirely new pull request.

src/roflibs/netlink/ctapdev.cpp
```diff
@@ -12,7 +12,7 @@
 #include "ctapdev.hpp"
 #include "roflibs/netlink/tap_manager.hpp"
 #include "roflibs/netlink/cpacketpool.hpp"
-#include "roflibs/netlink/clogging.hpp"
+#include <glog/logging.h>
```
octopusx: Same here, could be re-ordered.

PR #18, opened 2016-08-09
Reading rx_size and tx_size from /proc/sys/net/...

octopusx:
In cnetlink.cpp added file read statements to load the rx_size and tx_size buffer sizes from /proc/sys/net/core/rmem_max and /proc/sys/net/core/wmem_max respectively

### DIFF
--- a/src/roflibs/netlink/cnetlink.cpp
+++ b/src/roflibs/netlink/cnetlink.cpp
@@ -1,6 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#include <fstream>
 
 #include "cnetlink.hpp"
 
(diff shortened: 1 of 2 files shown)


src/roflibs/netlink/cnetlink.cpp
```diff
@@ -39,10 +39,30 @@ void cnetlink::init_caches() {
...
   }
 
-  // TODO read from /proc/sys/net/core/{r,w}mem_max
   int rx_size = 212992;
   int tx_size = 212992;
 
+  std::ifstream mem_max;
+  std::string output = "";
+
+  mem_max.open("/proc/sys/net/core/rmem_max");
```
toanju: can you create a function to do this, then you can reuse it below

src/roflibs/netlink/cnetlink.cpp
```diff
@@ -39,10 +39,30 @@ void cnetlink::init_caches() {
...
+  if (mem_max.is_open()) {
+    while (!mem_max.eof()) {
+       mem_max >> output;
```
toanju: you can read directly into an int, then you don't need atoi anymore

src/roflibs/netlink/cnetlink.hpp
```diff
@@ -7,6 +7,7 @@
 
 #include <deque>
 #include <exception>
+#include <fstream>
```
toanju: should go to cc file, since its not used in the header and it'll slow down compilation

src/roflibs/netlink/cnetlink.hpp
```diff
@@ -83,6 +83,8 @@ class cnetlink : public rofl::cthread_env, public rofcore::nbi {
 
   ~cnetlink() override;
 
+  int load_from_file(const std::string &path);
```
toanju: one last nitpick, please move this to the private section

PR #41, opened 2017-01-18
run tap io single threaded

toanju:
every tap interface currently creates a thread for its io. This is now
merged into a single thread for performance reasonns.

use override syntax for rofl interface

### DIFF
--- a/src/roflibs/netlink/ctapdev.cpp
+++ b/src/roflibs/netlink/ctapdev.cpp
@@ -17,34 +17,20 @@
 
 namespace rofcore {
 
-static inline void release_packets(std::deque<rofl::cpacket *> &q) {
-  for (auto i : q) {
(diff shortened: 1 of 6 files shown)


src/roflibs/netlink/ctapdev.cpp
```diff
@@ -17,34 +17,22 @@
...
 }
 
 void ctapdev::tap_open() {
   struct ifreq ifr;
   int rc;
 
   if (fd != -1) {
-    VLOG(1) << __FUNCTION__ << ": tapdev " << devname
-            << " is alredy open using fd=" << fd;
+    VLOG(1) << __FUNCTION__ << ": tapdev is alredy open using fd=" << fd;
```
DFritzsche: already

PR #66, opened 2017-11-03
Grpc merge

rubensfig:
Added gRPC capabilities by adding a connection between baseboxd and gRPC services, defined in basebox_api; 

It now exposes port 5000 for the gRPC server, and the responses it currently supports are to get the topology and statistics from baseboxd.


salt/roots/basebox/repos.sls
```diff
@@ -24,3 +24,15 @@ bisdn-rofl:
     - enabled_metadata: 1
     - require:
       - pkg: yum
+
+toanju-bisdn:
```
toanju: actually we could drop this one now

src/baseboxd.cpp
```diff
@@ -21,13 +22,23 @@ DEFINE_int32(port, 6653, "Listening port");
...
   bool running = true;
 
   if (!gflags::RegisterFlagValidator(&FLAGS_port, &validate_port)) {
     std::cerr << "Failed to register port validator" << std::endl;
     exit(1);
   }
+  ApiServer grpcConnector;
+
+  pthread_t tid;
+  pthread_create(&tid, NULL, startGRPC, &grpcConnector);
```
toanju: we should create the server after gflags has done it's magic

toanju: maybe we don't even need an additional thread here since the main loop does not really do anything

src/basebox_api.h
```diff
@@ -0,0 +1,36 @@
...
+#include <grpc++/server_builder.h>
+#include <grpc++/server_context.h>
+#include <grpc/grpc.h>
+
+#include "basebox_grpc_statistics.h"
+#include "basebox_grpc_topology.h"
+
+using rofl::openflow::cofport_stats_reply;
+using namespace openconfig_interfaces;
+using namespace ietf_network;
```
toanju: namespace pollution

src/basebox_grpc_statistics.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+#include <grpc++/grpc++.h>
+
+#include "services-definition.grpc.pb.h"
+#include "rofl/common/openflow/cofportstats.h"
+
+using namespace openconfig_interfaces;
+using namespace grpc;
```
toanju: namespace pollution

src/basebox_grpc_topology.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
+#include <grpc++/grpc++.h>
+
+#include "services-definition.grpc.pb.h"
+#include "rofl/common/openflow/cofportstats.h"
+
+using namespace ietf_network;
+using namespace grpc;
```
toanju: namespace pollution

DFritzsche:
general remark, could you split the PR into SAI and gRPC related changes?

PR #67, opened 2017-11-08
Added gRPC functionality

rubensfig:
Added gRPC capabilities by adding a connection between baseboxd and gRPC services, defined in basebox_api;

It now exposes port 5000 for the gRPC server, and the responses it currently supports are to get the topology and statistics from baseboxd.


src/basebox_api.cc
```diff
@@ -0,0 +1,44 @@
...
+  stats->addStatistics(nodeInfo, ports);
+
+  return 0;
+}
+
+int ApiServer::addLink(const std::string &nodeSrc, const std::string &portSrc,
+                       const std::string &nodeDst, const std::string &portDst) {
+  topology->addLink(nodeSrc, portSrc, nodeDst, portDst);
+  return 0;
+}
```
DFritzsche: we should think about try/catch in those functions

src/basebox_grpc_statistics.cc
```diff
@@ -0,0 +1,54 @@
...
+
+    counters->set_in_errors(port.second.get_rx_errors());
+    counters->set_in_fcs_errors(port.second.get_rx_crc_err());
+    counters->set_in_broadcast_pkts(port.second.get_rx_packets());
+    counters->set_in_discards(port.second.get_rx_dropped());
+    counters->set_in_octets(port.second.get_rx_bytes());
+  }
+}
+/*
+void NetworkImpl::fakeStatistic() {
```
DFritzsche: please delete

toanju:
everything should be included now in #68

PR #75, opened 2017-12-14
use of-dpa 3.0 group modifications

toanju:


src/of-dpa/controller.cpp
```diff
@@ -966,23 +966,23 @@ int controller::egress_bridge_port_vlan_add(uint32_t port, uint16_t vid,
     dpt.send_barrier_request(rofl::cauxid(0));
```
DFritzsche: lonely barrier, is that required?

src/of-dpa/controller.cpp
```diff
@@ -966,23 +966,23 @@ int controller::egress_bridge_port_vlan_add(uint32_t port, uint16_t vid,
...
+    auto l2_dom_it = l2_domain.find(vid);
+    if (l2_dom_it == l2_domain.end()) {
+      std::set<uint32_t> empty_set;
+      l2_dom_it = l2_domain.emplace(vid, empty_set).first;
+    }
+
+    // insert group_id to set
+    l2_dom_it->second.insert(group_id);
 
     // add new L2 flooding group
```
DFritzsche: this comment should also say "update" or something, see l. 1024

src/of-dpa/controller.cpp
```diff
@@ -1056,6 +1065,7 @@ int controller::egress_bridge_port_vlan_remove(uint32_t port,
     LOG(ERROR) << ": caught unknown exception: " << e.what();
```
DFritzsche: side note, why use ":" at the beginning? is this supposed to log `__FUNCTION__`?

src/of-dpa/controller.cpp
```diff
@@ -1056,6 +1065,7 @@ int controller::egress_bridge_port_vlan_remove(uint32_t port,
...
     rv = -EINVAL;
   }
+
```
DFritzsche: why add a blank line?

src/of-dpa/controller.cpp
```diff
@@ -966,23 +966,23 @@ int controller::egress_bridge_port_vlan_add(uint32_t port, uint16_t vid,
...
     // add new L2 flooding group
     dpt.send_group_mod_message(
         rofl::cauxid(0),
         fm_driver.enable_group_l2_flood(dpt.get_version(), vid, vid,
-                                        l2_domain[vid]));
+                                        l2_dom_it->second,
+                                        (l2_dom_it->second.size() != 1)));
     group_id = fm_driver.group_id_l2_flood(vid, vid);
 
     dpt.send_barrier_request(rofl::cauxid(0));
```
DFritzsche: this barrier should be placed directly after `enable_group_l2_flood`, because it's somewhat related to it, isn't it?

src/of-dpa/controller.cpp
toanju: not really lonely, since it's right after a group mod

toanju: not really lonely, since it's right after a group mod

src/of-dpa/controller.cpp
```diff
@@ -966,30 +966,33 @@ int controller::egress_bridge_port_vlan_add(uint32_t port, uint16_t vid,
...
         rofl::cauxid(0),
         fm_driver.enable_group_l2_flood(dpt.get_version(), vid, vid,
-                                        l2_domain[vid]));
+                                        l2_dom_it->second,
+                                        (l2_dom_it->second.size() != 1)));
+
     group_id = fm_driver.group_id_l2_flood(vid, vid);
 
-    dpt.send_barrier_request(rofl::cauxid(0));
+    if (l2_dom_it->second.size() == 1) { // send barrier on creation
```
DFritzsche: btw. is this `==1` good style or can it be deleted?

src/of-dpa/controller.cpp
```diff
@@ -1009,32 +1012,41 @@ int controller::egress_bridge_port_vlan_remove(uint32_t port,
...
+    l2_dom_it->second.erase(group_id);
 
     if (l2_domain[vid].size()) {
-      // add new L2 flooding group
+      // update L2 flooding group
       rofl::openflow::cofgroupmod gm = fm_driver.enable_group_l2_flood(
-          dpt.get_version(), vid, vid, l2_domain[vid]);
+          dpt.get_version(), vid, vid, l2_dom_it->second, true);
       dpt.send_group_mod_message(rofl::cauxid(0), gm);
       group_id = gm.get_group_id();
```
DFritzsche: we could get rid of creation of gm object if we would get group_id via `group_id = fm_driver.group_id_l2_flood(vid, vid);` as in the add function above

src/of-dpa/controller.cpp
```diff
@@ -1009,32 +1012,41 @@ int controller::egress_bridge_port_vlan_remove(uint32_t port,
...
 
       dpt.send_barrier_request(rofl::cauxid(0));
```
DFritzsche: this barrier seems to be not required, as the gm above has modify tag, see l. 990

src/of-dpa/controller.cpp
toanju: right, this was there for historic reasons...

src/of-dpa/controller.cpp
toanju: will check

PR #83, opened 2018-02-03
architectural rework to support additional ports

toanju:
Currently we just listen to events on ports we learned from the switch
api. To support additional ports we move the mapping of ifindex to
port_id (presented from the southbound api) to the tap_manager. Due to
this work cnetlink is no longer a singleton instance.

### DIFF
--- a/src/netlink/nl_bridge.cpp
+++ b/src/netlink/nl_bridge.cpp
@@ -14,12 +14,14 @@
 
 #include "sai.hpp"
 #include "nl_bridge.hpp"
+#include "tap_manager.hpp"
 
 namespace basebox {
 
-ofdpa_bridge::ofdpa_bridge(switch_interface *sw)
+ofdpa_bridge::ofdpa_bridge(switch_interface *sw,
+                           std::shared_ptr<tap_manager> tap_man)
     : bridge(nullptr), sw(sw), ingress_vlan_filtered(true),
-      egress_vlan_filtered(true) {}
+      egress_vlan_filtered(true), tap_man(tap_man) {}
 
 ofdpa_bridge::~ofdpa_bridge() {
   if (bridge)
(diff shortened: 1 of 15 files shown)


src/netlink/nl_bridge.cpp
```diff
@@ -339,10 +346,18 @@ void ofdpa_bridge::delete_interface(uint32_t port, rtnl_link *link) {
...
 }
 
-void ofdpa_bridge::add_mac_to_fdb(const uint32_t port, uint16_t vlan,
-                                  nl_addr *mac) {
+void ofdpa_bridge::add_neigh_to_fdb(rtnl_neigh *neigh) {
   assert(sw);
-  assert(mac);
+  assert(neigh);
+
+  uint32_t port = tap_man->get_port_id(rtnl_neigh_get_ifindex(neigh));
```
DFritzsche: this should be const (as on l. 378)

src/netlink/nl_l3.cpp
```diff
@@ -106,7 +109,7 @@ int nl_l3::add_l3_termination(struct rtnl_addr *a) {
   }
 
   int ifindex = rtnl_addr_get_ifindex(a);
-  int port_id = cnetlink::get_instance().get_port_id(ifindex);
+  int port_id = tap_man->get_port_id(ifindex);
```
DFritzsche: consider making this port_id const, everything that comes out of a get function should be const in my opinion (not a complain)

src/netlink/tap_manager.hpp
```diff
@@ -80,11 +80,41 @@ class tap_manager final {
...
 private:
   tap_manager(const tap_manager &other) = delete; // non construction-copyable
   tap_manager &operator=(const tap_manager &) = delete; // non copyable
 
-  std::map<uint32_t, ctapdev *> devs;
+  std::map<uint32_t, ctapdev *> tap_devs; // southbound id:tap_device
+  mutable std::mutex rp_mutex;
+  std::map<std::string, uint32_t> tap_names;
+
+  std::map<int, uint32_t> ifindex_to_id;
```
DFritzsche: so, index should not be smaller than 0, so why not use `uint` instead?

src/netlink/nl_bridge.cpp
toanju: First we have to change the API call (l2_addr_add) that is called with this value otherwise this does not work. This PR does not takle the API though.

src/netlink/tap_manager.hpp
toanju: we use the same type for the ifindex as used by the kernel

src/netlink/nl_bridge.cpp
toanju: could be, I don't have a strong opinion here since it's a basic type and not passed as a reference/pointer.

src/netlink/nl_l3.cpp
toanju: why should this be related to a get function?

PR #85, opened 2018-02-06
tap_manager: do not lock during tap creation

toanju:
during tap creation we still receive notifications from cnetlink. This
cannot be block since netlink messages might be dropped. fixes f1d021e

### DIFF
--- a/src/netlink/tap_manager.cpp
+++ b/src/netlink/tap_manager.cpp
@@ -153,18 +153,32 @@ tap_manager::~tap_manager() { destroy_tapdevs(); }
 int tap_manager::create_tapdev(uint32_t port_id, const std::string &port_name,
                                switch_callback &cb) {
   int r = 0;
-  std::lock_guard<std::mutex> lock{rp_mutex};
+  bool dev_exists = false;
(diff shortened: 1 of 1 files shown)


src/netlink/tap_manager.cpp
```diff
@@ -153,18 +153,31 @@ tap_manager::~tap_manager() { destroy_tapdevs(); }
...
+      dev_name_exists = true;
+  }
+
+  if (!dev_exists && !dev_name_exists) {
     // create a new tap device
 
     ctapdev *dev;
     try {
       dev = new ctapdev(port_name);
       tap_devs.insert(std::make_pair(port_id, dev));
```
DFritzsche: operation on tap_devs is not locked, but above it is

toanju: good catch, so I will remove it above as well since this structure does not need to be locked.

DFritzsche: so l. 164/165 can be removed from lock as well, also this would improve code clarity

toanju: done

PR #86, opened 2018-02-07
support multiple devices

toanju:
to support different link types cnetlink is refactored to simplify
addition of new link types like vxlan

### DIFF
--- a/src/netlink/cnetlink.cpp
+++ b/src/netlink/cnetlink.cpp
@@ -23,12 +23,23 @@
 #define ROUTE_CAST(obj) reinterpret_cast<struct rtnl_route *>(obj)
 #define ADDR_CAST(obj) reinterpret_cast<struct rtnl_addr *>(obj)
 
+#define lt_names "unknown", "bridge", "tun"
+
 namespace basebox {
 
 cnetlink::cnetlink(switch_interface *swi, std::shared_ptr<tap_manager> tap_man)
     : swi(swi), thread(this), caches(NL_MAX_CACHE, nullptr), tap_man(tap_man),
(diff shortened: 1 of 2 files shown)


src/netlink/cnetlink.cpp
```diff
@@ -23,12 +23,24 @@
...
       bridge(nullptr), nl_proc_max(10), running(false), rfd_scheduled(false),
-      l3(swi, tap_man, this) {
+      l3(swi, tap_man, this), lt2names{lt_names} {
+
+  assert(lt2names.size() == LT_MAX);
+
+  enum link_type lt = LT_UNKNOWN;
+  for (const auto &n : lt2names) {
+    assert(lt != LT_MAX);
+    LOG(INFO) << n;
```
DFritzsche: a wild log message

src/netlink/cnetlink.cpp
```diff
@@ -23,12 +23,24 @@
 #define ROUTE_CAST(obj) reinterpret_cast<struct rtnl_route *>(obj)
 #define ADDR_CAST(obj) reinterpret_cast<struct rtnl_addr *>(obj)
 
+#define lt_names "unknown", "bridge", "tun"
```
DFritzsche: why no integrate this into `enum link_type`?

toanju: these are the name that besides `unknow` match the result of rtnl_link_get_type (i.e. the identifier/kind of the netlink operations).

src/netlink/cnetlink.cpp
toanju: wild thing... yup that has to go away

PR #94, opened 2018-03-05
clear l2_domain in case of dpt_closed

DFritzsche:
* clear l2_domain explicitely when dpt_closed

### DIFF
--- a/src/of-dpa/controller.cpp
+++ b/src/of-dpa/controller.cpp
@@ -50,6 +50,12 @@ void controller::handle_dpt_close(const rofl::cdptid &dptid) {
 
   std::deque<nbi::port_notification_data> ntfys;
   try {
+    {
+      std::lock_guard<std::mutex> lock(l2_domain_mutex);
+      // clear local state, all ports are gone
+      l2_domain.clear();
+    }
+
     // TODO check dptid and dptid?
     rofl::crofdpt &dpt = set_dpt(dptid, true);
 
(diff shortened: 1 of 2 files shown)


src/of-dpa/controller.cpp
```diff
@@ -50,6 +50,9 @@ void controller::handle_dpt_close(const rofl::cdptid &dptid) {
 
   std::deque<nbi::port_notification_data> ntfys;
   try {
+    // clear local state, all ports are gone
+    l2_domain.clear();
```
toanju: unfortunately we have to add a lock now for the l2_domain

src/of-dpa/controller.cpp
```diff
@@ -964,24 +970,27 @@ int controller::egress_bridge_port_vlan_add(uint32_t port, uint16_t vid,
...
+      auto l2_dom_it = l2_domain.find(vid);
+      if (l2_dom_it == l2_domain.end()) {
+        std::set<uint32_t> empty_set;
+        l2_dom_it = l2_domain.emplace(vid, empty_set).first;
+      }
 
-    // insert group_id to set
-    l2_dom_it->second.insert(fm_driver.group_id_l2_interface(port, vid));
+      // insert group_id to set
+      l2_dom_it->second.insert(fm_driver.group_id_l2_interface(port, vid));
```
toanju: the lock could be already release here and the domain size stored in a local int

src/of-dpa/controller.cpp
```diff
@@ -1007,37 +1016,39 @@ int controller::egress_bridge_port_vlan_remove(uint32_t port,
...
-      // remove L2 flooding group
-      dpt.send_flow_mod_message(
-          rofl::cauxid(0),
-          fm_driver.remove_bridging_dlf_vlan(dpt.get_version(), vid));
-      dpt.send_barrier_request(rofl::cauxid(0));
-      dpt.send_group_mod_message(
-          rofl::cauxid(0),
-          fm_driver.disable_group_l2_flood(dpt.get_version(), vid, vid));
+      // remove group_id from set
+      l2_dom_it->second.erase(fm_driver.group_id_l2_interface(port, vid));
```
toanju: Let's keep the scope of the lock minimal and thus release the lock after the erasure here. Despite we have to copy the set.

src/of-dpa/controller.cpp
DFritzsche: `l2_dom_it->second` is also used below, this needs a local set, are you sure?

PR #101, opened 2018-04-11
Refactoring

toanju:
a lot of refactoring to support future vlan, vxlan and l3 improvements

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -44,8 +44,8 @@ SOURCES_NETLINK = \
 	netlink/nbi_impl.hpp \
 	netlink/nl_bridge.cpp \
 	netlink/nl_bridge.hpp \
-	netlink/nl_l3.hpp \
 	netlink/nl_l3.cpp \
+	netlink/nl_l3.hpp \
 	netlink/nl_obj.cpp \
 	netlink/nl_obj.hpp \
 	netlink/nl_output.cpp \
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -42,6 +42,8 @@ SOURCES_NETLINK = \
 	netlink/ctapdev.hpp \
 	netlink/nbi_impl.cpp \
 	netlink/nbi_impl.hpp \
+	netlink/netlink-utils.cpp \
+	netlink/netlink-utils.hpp \
 	netlink/nl_bridge.cpp \
 	netlink/nl_bridge.hpp \
 	netlink/nl_l3.cpp \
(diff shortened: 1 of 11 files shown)


src/netlink/cnetlink.cpp
```diff
@@ -16,31 +16,18 @@
...
-  enum link_type lt = LT_UNKNOWN;
-  for (const auto &n : lt2names) {
-    assert(lt != LT_MAX);
-    kind2lt.emplace(n, lt);
-    lt = static_cast<enum link_type>(lt + 1);
-  }
+cnetlink::cnetlink(std::shared_ptr<tap_manager> tap_man)
+    : swi(nullptr), thread(this), caches(NL_MAX_CACHE, nullptr),
+      tap_man(tap_man), bridge(nullptr), nl_proc_max(10), running(false),
+      rfd_scheduled(false), l3(new nl_l3(tap_man, this)) {
```
DFritzsche: where is the 'delete' for nl_l3?

src/netlink/netlink-utils.hpp
```diff
@@ -0,0 +1,32 @@
+#pragma once
+
+#include <cassert>
+#include <cinttypes>
+#include <deque>
```
DFritzsche: I think deque and assert are not used in this file

src/Makefile.am
```diff
@@ -42,10 +42,13 @@ SOURCES_NETLINK = \
 	netlink/ctapdev.hpp \
 	netlink/nbi_impl.cpp \
 	netlink/nbi_impl.hpp \
+	netlink/netlink-utils.cpp \
+	netlink/netlink-utils.hpp \
 	netlink/nl_bridge.cpp \
 	netlink/nl_bridge.hpp \
-	netlink/nl_l3.hpp \
+	netlink/nl_hasing.hpp \
```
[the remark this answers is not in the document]
toanju: hm, thought I fixed it here as well. Thanks!

src/netlink/cnetlink.cpp
toanju: the shared_ptr will do this

src/netlink/netlink-utils.hpp
toanju: will check

src/Makefile.am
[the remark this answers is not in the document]
toanju: For compilation only this is not for interest. It would not be able to build a dist archive.

PR #104, opened 2018-04-20
L2 ageing

toanju:
bridging entries are now no longer timing out despite having traffic on them

### DIFF
--- a/src/netlink/cnetlink.cpp
+++ b/src/netlink/cnetlink.cpp
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 #include <cassert>
 #include <cstring>
+#include <exception>
 #include <fstream>
 #include <glog/logging.h>
 #include <iterator>
@@ -29,12 +30,6 @@ cnetlink::cnetlink(std::shared_ptr<tap_manager> tap_man)
       tap_man(tap_man), bridge(nullptr), nl_proc_max(10), running(false),
       rfd_scheduled(false), l3(new nl_l3(tap_man, this)) {
 
-  sock = nl_socket_alloc();
-  if (NULL == sock) {
-    LOG(FATAL) << "cnetlink: failed to create netlink socket" << __FUNCTION__;
-    throw eNetLinkCritical(__FUNCTION__);
-  }
-
   try {
     thread.start("netlink");
     init_caches();
@@ -47,7 +42,7 @@ cnetlink::~cnetlink() {
   thread.stop();
   delete bridge;
   destroy_caches();
-  nl_socket_free(sock);
+  nl_socket_free(sock_mon);
 }
 
 int cnetlink::load_from_file(const std::string &path) {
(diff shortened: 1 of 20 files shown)


src/netlink/cnetlink.cpp
```diff
@@ -282,6 +300,114 @@ void cnetlink::nl_cb_v2(struct nl_cache *cache, struct nl_object *old_obj,
...
+      VLOG(2) << __FUNCTION__ << ": ifindex=" << ifindex
+              << ", bridge=" << bridge << ", br_link=" << OBJ_CAST(br_link);
+
+      if (br_link) {
+        // learn the source mac
+        bridge->learn_source_mac(br_link, p.pkt);
+      }
+    }
+
+    VLOG(2) << __FUNCTION__ << ": send pkt " << p.pkt
```
DFritzsche: not going to stress this, but technically speaking the packets are not sent to the kernel, just queued into the tap_man

src/of-dpa/controller.cpp
```diff
@@ -561,14 +462,23 @@ int controller::l2_addr_remove_all_in_vlan(uint32_t port,
...
 int controller::l2_addr_add(uint32_t port, uint16_t vid,
-                            const rofl::cmacaddr &mac, bool filtered) noexcept {
+                            const rofl::caddress_ll &mac, bool filtered,
+                            bool permanent) noexcept {
   int rv = 0;
   try {
     rofl::crofdpt &dpt = set_dpt(dptid, true);
+
+    if (!permanent)
+      fm_driver.set_idle_timeout(300);
```
DFritzsche: should the 300 be configurable?

src/of-dpa/controller.cpp
```diff
@@ -127,28 +122,8 @@ void controller::handle_packet_in(rofl::crofdpt &dpt, const rofl::cauxid &auxid,
...
-  case OFDPA_FLOW_TABLE_ID_ACL_POLICY:
-    send_packet_in_to_cpu(dpt, msg);
-    break;
-  default:
-    LOG(WARNING) << __FUNCTION__ << ": unexpected packet-in from table "
-                 << msg.get_table_id() << ". Dropping packet.";
-    break;
-  }
+  // all packets go up
+  send_packet_in_to_cpu(dpt, msg);
```
DFritzsche: sometimes its cpu, sometimes kernel, are those the same?

src/of-dpa/controller.cpp
```diff
@@ -158,27 +133,20 @@ void controller::handle_flow_removed(rofl::crofdpt &dpt,
...
-  }
-#endif
-
   switch (msg.get_table_id()) {
   case OFDPA_FLOW_TABLE_ID_BRIDGING:
     handle_bridging_table_rm(dpt, msg);
     break;
   default:
+    LOG(WARNING) << __FUNCTION__ << ": unhandled flow removal in table_id="
+                 << msg.get_table_id();
```
DFritzsche: table_id should be casted to int to get useful logging

src/netlink/tap_manager.cpp
```diff
@@ -163,35 +27,41 @@ int tap_manager::create_tapdev(uint32_t port_id, const std::string &port_name,
...
       int fd = dev->get_fd();
 
       LOG(INFO) << __FUNCTION__ << ": port_id=" << port_id
                 << " portname=" << port_name << " fd=" << fd << " ptr=" << dev;
 
-      // start reading from port
-      io.register_tap(fd, port_id, cb);
+      {
+        std::lock_guard<std::mutex> lock{tn_mutex};
+        tap_names2fds.emplace(std::make_pair(port_name, fd));
```
DFritzsche: tap_names2fds is neither cleared nor erased

src/netlink/cnetlink.cpp
```diff
@@ -282,6 +300,114 @@ void cnetlink::nl_cb_v2(struct nl_cache *cache, struct nl_object *old_obj,
...
+  std::deque<nl_pkt_in> _packet_in;
+
+  {
+    std::lock_guard<std::mutex> scoped_lock(pi_mutex);
+    _packet_in.swap(packet_in);
+  }
+
+  for (int cnt = 0; cnt < nl_proc_max && _packet_in.size() && running; cnt++) {
+    auto p = _packet_in.front();
+    int ifindex = tap_man->get_ifindex(p.port_id);
```
DFritzsche: so, that's an `ifi` then ;)

src/netlink/cnetlink.cpp
```diff
@@ -25,10 +25,20 @@
...
+cnetlink::cnetlink()
     : swi(nullptr), thread(this), caches(NL_MAX_CACHE, nullptr),
-      tap_man(tap_man), bridge(nullptr), nl_proc_max(10), running(false),
-      rfd_scheduled(false), l3(new nl_l3(tap_man, this)) {
+      bridge(nullptr), nl_proc_max(10), running(false), rfd_scheduled(false),
+      l3(new nl_l3(this)) {
+
+  sock_tx = nl_socket_alloc();
+  if (sock_tx == nullptr) {
+    LOG(FATAL) << "cnetlink: failed to create netlink socket" << __FUNCTION__;
```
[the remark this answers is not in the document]
toanju: historic reasons (i.e. did not touch that part)

src/netlink/cnetlink.cpp
```diff
@@ -282,6 +300,114 @@ void cnetlink::nl_cb_v2(struct nl_cache *cache, struct nl_object *old_obj,
...
+    tap_man->enqueue(p.fd, p.pkt);
+
+    _packet_in.pop_front();
+  }
+
+  int size = _packet_in.size();
+  if (size) {
+    std::lock_guard<std::mutex> scoped_lock(pi_mutex);
+    std::copy(make_move_iterator(_packet_in.begin()),
+              make_move_iterator(packet_in.end()),
```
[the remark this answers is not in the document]
toanju: good catch, thanks

src/netlink/cnetlink.cpp
```diff
@@ -282,6 +300,114 @@ void cnetlink::nl_cb_v2(struct nl_cache *cache, struct nl_object *old_obj,
...
+              std::back_inserter(packet_in));
```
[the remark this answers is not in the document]
toanju: true, completely forgot about this, thanks

src/netlink/cnetlink.cpp
toanju: ah, one line below... fixed

src/of-dpa/controller.cpp
toanju: could, but it's pretty much fixed in the kernel as well hence I would do this as a follow up (later)

src/of-dpa/controller.cpp
toanju: It was just kernel in the 'comment' (fixed above). 'To CPU' points into the direction the packets go which in our case eventually will be the Linux kernel. Thus both would be correct though.

src/of-dpa/controller.cpp
toanju: unsigned ;)

src/netlink/tap_manager.cpp
toanju: good catch, thanks

src/netlink/cnetlink.cpp
toanju: :rofl:

src/netlink/tap_manager.cpp
DFritzsche: we should add a `tap_names2fds.clear()` to here: https://github.com/toanju/basebox/blob/b985724583959f13a72ac9cad8e3e110d99635aa/src/netlink/tap_manager.cpp#L118

src/of-dpa/controller.cpp
DFritzsche: well, I like having one name for one thing, that's all

src/netlink/tap_manager.cpp
toanju: solved slightly differntly

PR #108, opened 2018-05-17
Added way to sync the previously existing lo interfaces

rubensfig:
baseboxd now synchronizes the previously configured IP addresses on the loopback interface.

### DIFF
--- a/src/netlink/cnetlink.cpp
+++ b/src/netlink/cnetlink.cpp
@@ -31,7 +31,7 @@ namespace basebox {
 cnetlink::cnetlink()
     : swi(nullptr), thread(this), caches(NL_MAX_CACHE, nullptr),
       bridge(nullptr), nl_proc_max(10), running(false), rfd_scheduled(false),
-      vlan(new nl_vlan(this)), l3(new nl_l3(vlan, this)) {
+      lo_processed(false), vlan(new nl_vlan(this)), l3(new nl_l3(vlan, this)) {
 
   sock_tx = nl_socket_alloc();
   if (sock_tx == nullptr) {
@@ -304,12 +304,18 @@ void cnetlink::handle_wakeup(rofl::cthread &thread) {
     do_wakeup = true;
   }
 
+  if (swi && !lo_processed) {
+    config_lo_addr();
+    lo_processed = false;
+  }
+
   if (do_wakeup || nl_objs.size()) {
     this->thread.wakeup();
   }
 }
 
 void cnetlink::handle_read_event(rofl::cthread &thread, int fd) {
+
   if (fd == nl_cache_mngr_get_fd(mngr)) {
     int rv = nl_cache_mngr_data_ready(mngr);
     VLOG(1) << __FUNCTION__ << ": #processed=" << rv;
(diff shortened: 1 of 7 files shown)


src/netlink/nl_l3.cpp
```diff
@@ -87,11 +87,13 @@ int nl_l3::add_l3_addr(struct rtnl_addr *a) {
   }
 
   bool is_loopback = (rtnl_link_get_flags(link) & IFF_LOOPBACK);
+
```
toanju: weird empty line

src/netlink/nl_l3.cpp
```diff
@@ -87,11 +87,13 @@ int nl_l3::add_l3_addr(struct rtnl_addr *a) {
...
   int ifindex = 0;
   uint16_t vid = vlan->get_vid(link);
 
   // XXX TODO split this into several functions
   if (!is_loopback) {
+
```
toanju: weird empty line

src/netlink/cnetlink.cpp
```diff
@@ -42,6 +42,7 @@ cnetlink::cnetlink()
 
   set_nl_socket_buffer_sizes(sock_tx);
 
+  lo_processed = false;
```
toanju: even though this is a plain old data type let's move this to the initializer list above

src/netlink/cnetlink.cpp
```diff
@@ -1083,4 +1090,44 @@ int cnetlink::handle_port_status_events() {
   return size;
 }
 
+int cnetlink::config_lo_addr() noexcept {
+  struct rtnl_link *lo_link;
+  std::list<struct rtnl_addr *> lo_addr;
+
+  std::unique_ptr<struct rtnl_link, void (*)(rtnl_link *)> filter(
```
toanju: instead of `filter` and `filter2` let's use `link_filter` and `addr_filter`

src/netlink/cnetlink.cpp
```diff
@@ -1083,4 +1090,44 @@ int cnetlink::handle_port_status_events() {
   return size;
 }
 
+int cnetlink::config_lo_addr() noexcept {
+  struct rtnl_link *lo_link;
+  std::list<struct rtnl_addr *> lo_addr;
+
```
toanju: empty line could be remove

src/netlink/cnetlink.cpp
```diff
@@ -1083,4 +1090,44 @@ int cnetlink::handle_port_status_events() {
...
+  std::unique_ptr<struct rtnl_link, void (*)(rtnl_link *)> filter(
+      rtnl_link_alloc(), &rtnl_link_put);
+
```
toanju: empty line could be remove hence all new variables are in one block

src/netlink/cnetlink.cpp
```diff
@@ -1083,4 +1090,44 @@ int cnetlink::handle_port_status_events() {
...
+        std::list<struct rtnl_addr *> *add_list =
+            static_cast<std::list<struct rtnl_addr *> *>(arg);
+        add_list->emplace_back(ADDR_CAST(obj));
+      },
+      &lo_addr);
+
+  for (auto addr : lo_addr) {
+    if (l3->add_l3_addr(addr) < 0)
+      return 1;
```
toanju: we should return only negative values in case of an error I guess this would be a `-EINVAL`

http://man7.org/linux/man-pages/man3/errno.3.html

src/netlink/cnetlink.cpp
```diff
@@ -1083,4 +1090,44 @@ int cnetlink::handle_port_status_events() {
...
+      rtnl_link_alloc(), &rtnl_link_put);
+
+  std::unique_ptr<struct rtnl_addr, void (*)(rtnl_addr *)> filter2(
+      rtnl_addr_alloc(), &rtnl_addr_put);
+
+  rtnl_link_set_flags(filter.get(), IFF_LOOPBACK);
+
+  // search interface cache by loopback filter, is this really needed if only
+  // one??
+  nl_cache_foreach_filter(caches[NL_LINK_CACHE], OBJ_CAST(filter.get()),
```
toanju: seems like the loopback is always 1, hence we could shorten this and use link_lookup

https://elixir.bootlin.com/linux/latest/source/include/net/flow.h#L23 https://elixir.bootlin.com/linux/latest/source/drivers/net/loopback.c#L218

toanju: ps: and I was wrong, there can be only one lo interface

src/netlink/netlink-utils.hpp
```diff
@@ -34,6 +34,6 @@ enum link_type {
 enum link_type kind_to_link_type(const char *type) noexcept;
 
 void get_bridge_ports(int br_ifindex, struct nl_cache *link_cache,
-                      std::deque<rtnl_link *> *list) noexcept;
+                      std::deque<rtnl_link *> *link_list) noexcept;
```
toanju: should have been in a separate PR, but works for me here

src/netlink/nl_l3.cpp
```diff
@@ -124,11 +124,13 @@ int nl_l3::add_l3_addr(struct rtnl_addr *a) {
 
   rv = sw->l3_unicast_host_add(ipv4_dst,
                                0); // TODO likely move this to separate entity
+#if 0
```
toanju: last nitpick, let's keep this error message for now

PR #110, opened 2018-05-28
Fix bug with loopback interfaces netmask

rubensfig:


src/netlink/cnetlink.cpp
```diff
@@ -1096,11 +1096,13 @@ int cnetlink::config_lo_addr() noexcept {
...
   rtnl_addr_set_ifindex(addr_filter.get(), 1);
   rtnl_addr_set_family(addr_filter.get(), AF_INET);
+  rtnl_addr_set_prefixlen(addr_filter.get(), 32);
 
   nl_cache_foreach_filter(
       caches[NL_ADDR_CACHE], OBJ_CAST(addr_filter.get()),
       [](struct nl_object *obj, void *arg) {
-        VLOG(3) << __FUNCTION__ << " : found configured loopback " << obj;
+        VLOG(3) << __FUNCTION__ << " : found configured loopback "
+                << OBJ_CAST(obj);
```
toanju: no need to cast here

src/netlink/cnetlink.cpp
```diff
@@ -1111,7 +1113,7 @@ int cnetlink::config_lo_addr() noexcept {
 
   for (auto addr : lo_addr) {
     if (l3->add_l3_addr(addr) < 0)
-      return -EINVAL;
+      VLOG(2) << __FUNCTION__ << " : failed to configure addr";
```
toanju: What's the reason for dropping the return value?

src/netlink/nl_l3.cpp
```diff
@@ -119,14 +119,21 @@ int nl_l3::add_l3_addr(struct rtnl_addr *a) {
   }
 
   // get v4 dst (local v4 addr)
+  auto i = rtnl_addr_get_prefixlen(a);
```
toanju: please rename to e.g. prefixlen

PR #134, opened 2018-09-28
systemd-service: remove network-online.target dependency

toanju:
This just causes trouble and can be omitted


hwoesner:
Oh, I see, at least in the case of having baseboxd on the box this may potentially be harmful.
In the other case, there is still no strict dependency needed.

PR #137, opened 2018-10-08
Added basic IPv6 neighbor and address functions

rubensfig:


src/netlink/nl_l3.cpp
```diff
@@ -176,6 +182,96 @@ int nl_l3::add_l3_addr(struct rtnl_addr *a) {
...
+  VLOG(1) << __FUNCTION__ << ": added addr " << OBJ_CAST(a);
+
+  return rv;
+}
+
+int nl_l3::add_lo_addr_v6(struct rtnl_addr *a) {
+  int rv = 0;
+  auto addr = rtnl_addr_get_local(a);
+
+  auto p = nl_addr_alloc(255);
```
toanju: a length of 16 should be sufficient

src/netlink/nl_l3.cpp
```diff
@@ -325,9 +422,23 @@ int nl_l3::add_l3_neigh(struct rtnl_neigh *n) {
   }
 
   addr = rtnl_neigh_get_dst(n);
-  rofl::caddress_in4 ipv4_dst = libnl_in4addr_2_rofl(addr);
+  if (family == AF_INET) {
+    rofl::caddress_in4 ipv4_dst = libnl_in4addr_2_rofl(addr);
+    rv = sw->l3_unicast_host_add(ipv4_dst, l3_interface_id);
+  } else {
+
+    auto p = nl_addr_alloc(255);
```
toanju: same here

src/netlink/nl_l3.cpp
```diff
@@ -801,4 +920,15 @@ void nl_l3::get_neighbours_of_route(rtnl_route *route, nh_lookup_params *p) {
       p);
 }
 
+bool nl_l3::is_link_local_address(const struct nl_addr *addr) {
+  auto p = nl_addr_alloc(255);
```
toanju: again length 16

PR #142, opened 2018-10-19
build toolchain: replace autotools with meson

toanju:
./configure && make will still run but they are a wrapper for meson and
ninja. Hence the preferred way to build is by calling
  meson build && ninja -C build


rubensfig:
Should meson also be listed as a dependency for building?

rubensfig:
The salt files should also be updated, removing the autotools installer, for the Vagrant environment.

toanju:
thanks, updated

PR #144, opened 2018-10-22
Port removal must not crash baseboxd

rubensfig:


toanju:
closing and opening to trigger travis

src/netlink/tap_manager.hpp
```diff
@@ -65,8 +65,11 @@ class tap_manager final {
 
   int get_fd(uint32_t port_id) const noexcept;
 
-  void tap_dev_ready(int ifindex, const std::string &name);
-  void tap_dev_removed(int ifindex);
+  void tapdev_ready(int ifindex, const std::string &name);
+
+  // access from northbound (cnetlink)
+  int remove_tapdev(int ifindex);
```
toanju: I think this should still be named `tapdev_removed` since the device was already removed and this is the notification via netlink. This way it would be in line with `tapdev_ready` which should be listed as well below comment `access from northbound`

src/netlink/cnetlink.cpp
```diff
@@ -943,7 +943,13 @@ void cnetlink::link_deleted(rtnl_link *link) noexcept {
     }
     break;
   case LT_TUN:
-    tap_man->tap_dev_removed(rtnl_link_get_ifindex(link));
+    if (tap_man->remove_tapdev(rtnl_link_get_ifindex(link)) == -EINVAL) {
+      tap_man->recreate_tapdev(rtnl_link_get_ifindex(link),
```
toanju: would it be possible to do this from the tap manager?

toanju: we could pass as well the whole link structure

PR #146, opened 2018-10-25
bond: add single port bond wrapper

toanju:
* update link_type handling
* handle single bond interface


rubensfig:
Should we also update the README, with the version of libnl?

rubensfig:
Sorry, accidentally closed the PR

toanju:
I'm currently thinking about having this lib as a submodule since we depend on the master tree here.

PR #156, opened 2018-11-07
fix example scripts

DFritzsche:
* fixes several errors in the example configuration scripts


examples/bridging/01-controller
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-
+
```
toanju:
```suggestion

```

PR #160, opened 2018-11-14
Updating neighbors enabled

rubensfig:


src/netlink/nl_l3.cpp
```diff
@@ -593,7 +591,47 @@ int nl_l3::update_l3_neigh(struct rtnl_neigh *n_old, struct rtnl_neigh *n_new) {
...
+      l3_interface_id = it->second.l3_interface_id;
+
+    rv = sw->l3_egress_update(port_id, vid, libnl_lladdr_2_rofl(s_mac),
+                              libnl_lladdr_2_rofl(n_ll_new), &l3_interface_id);
+    if (rv < 0) {
+      VLOG(2) << __FUNCTION__ << ": failed to add neighbor";
+      return -EINVAL;
+    }
+
+    l3_interface_mapping.erase(l3_if_old_tuple);
```
toanju: here the iterator should be used to not do the lookup again

src/of-dpa/controller.cpp
```diff
@@ -714,6 +714,7 @@ int controller::l3_egress_create(uint32_t port, uint16_t vid,
   int rv = 0;
   uint32_t _egress_interface_id;
   bool increment = false;
+  bool modify = false;
```
toanju: I don't think we need to spend a variable on this since modify is just passed.

src/of-dpa/controller.cpp
```diff
@@ -751,6 +752,35 @@ int controller::l3_egress_create(uint32_t port, uint16_t vid,
   return rv;
 }
 
+int controller::l3_egress_update(uint32_t port, uint16_t vid,
+                                 const rofl::caddress_ll &src_mac,
+                                 const rofl::caddress_ll &dst_mac,
+                                 uint32_t *l3_interface_id) noexcept {
+  int rv = 0;
+  bool modify = true;
```
toanju: I don't think we need to spend a variable on this since modify is just passed.

src/netlink/nl_l3.cpp
```diff
@@ -593,7 +591,49 @@ int nl_l3::update_l3_neigh(struct rtnl_neigh *n_old, struct rtnl_neigh *n_new) {
...
+                        libnl_lladdr_2_rofl(n_ll_new));
+
+    // Obtain l3_interface_id
+    auto it_range = l3_interface_mapping.equal_range(l3_if_old_tuple);
+    uint32_t l3_interface_id = 0;
+    for (auto it = it_range.first; it != it_range.second; ++it) {
+      l3_interface_id = it->second.l3_interface_id;
+
+      rv =
+          sw->l3_egress_update(port_id, vid, libnl_lladdr_2_rofl(s_mac),
```
toanju: This will update all egress l3 interfaces even if there is a hash collision. There should be rather a check:
```
if (l3_if_old_tuple == it->first) { 
/*actual l3_interface found */ 
}
```
Furthermore the case where no matching tuple is retrieved is not covered.

src/netlink/nl_l3.cpp
```diff
@@ -593,7 +591,55 @@ int nl_l3::update_l3_neigh(struct rtnl_neigh *n_old, struct rtnl_neigh *n_new) {
...
+          return -EINVAL;
+        }
+
+        l3_interface_mapping.erase(it_range.first, it_range.second);
+        l3_interface_mapping.emplace(
+            std::make_pair(l3_if_new_tuple, l3_interface(l3_interface_id)));
+
+      } else {
+        VLOG(2) << __FUNCTION__ << ": skipping interface";
+        continue;
```
toanju: no continue needed here

src/netlink/nl_l3.cpp
```diff
@@ -593,7 +591,55 @@ int nl_l3::update_l3_neigh(struct rtnl_neigh *n_old, struct rtnl_neigh *n_new) {
...
+
+    auto l3_if_old_tuple =
+        std::make_tuple(port_id, vid, libnl_lladdr_2_rofl(s_mac),
+                        libnl_lladdr_2_rofl(n_ll_old));
+    auto l3_if_new_tuple =
+        std::make_tuple(port_id, vid, libnl_lladdr_2_rofl(s_mac),
+                        libnl_lladdr_2_rofl(n_ll_new));
+
+    // Obtain l3_interface_id
+    auto it_range = l3_interface_mapping.equal_range(l3_if_old_tuple);
```
toanju: Can it happen that the l3_if_old_tuple is not found?

src/netlink/nl_l3.cpp
```diff
@@ -593,7 +591,60 @@ int nl_l3::update_l3_neigh(struct rtnl_neigh *n_old, struct rtnl_neigh *n_new) {
...
+          return -EINVAL;
+        }
+        break;
+
+      } else {
+        VLOG(2) << __FUNCTION__ << ": skipping interface";
+      }
+    }
+    l3_interface_mapping.erase(it->first);
+    l3_interface_mapping.emplace(
```
toanju: I think we have to cover one more case which is when the new_tuple is already existing.

PR #161, opened 2018-11-14
Bug adding neighbors when they are showing as INCOMPLETE

rubensfig:


src/netlink/nl_l3.cpp
```diff
@@ -402,6 +402,11 @@ int nl_l3::add_l3_neigh_egress(struct rtnl_neigh *n,
     return -EINVAL;
   }
 
+  if (state == NUD_INCOMPLETE) {
```
toanju: let's add a switch case block here and converge with the invalid above

PR #163, opened 2018-12-05
fix: race condition during port creation; all ports up

rubensfig:


src/netlink/cnetlink.cpp
```diff
@@ -1089,93 +1089,95 @@ void cnetlink::port_status_changed(uint32_t port_no,
...
-        _pc_retry.push_back(change);
-      } else {
-        LOG(ERROR) << __FUNCTION__
-                   << ": no ifindex of port_id=" << std::get<0>(change)
-                   << " found";
-        // XXX TODO resync caches?
-      }
-      continue;
-    }
+    _pc_changes = port_status_changes;
```
DFritzsche: assign the `change` object here, since you are only using the first element

src/netlink/cnetlink.cpp
```diff
@@ -1089,93 +1089,95 @@ void cnetlink::port_status_changed(uint32_t port_no,
...
+  }
 
-    VLOG(1) << __FUNCTION__ << ": update link with ifindex=" << ifindex
-            << " port_id=" << std::get<0>(change) << " status=" << std::hex
-            << std::get<1>(change) << std::dec << ") ";
+  change = _pc_changes.front();
+  size = _pc_changes.size();
```
DFritzsche: when doign the changes above, assign size within the lock, that should still be faster than copying the entire deque

src/netlink/cnetlink.cpp
```diff
@@ -1089,93 +1089,95 @@ void cnetlink::port_status_changed(uint32_t port_no,
 }
 
 int cnetlink::handle_port_status_events() {
+  if (!port_status_changes.size())
```
DFritzsche: this has no lock and could cause trouble, why not get the size within the lock (then exit if 0) as described in the comment below

src/netlink/cnetlink.cpp
```diff
@@ -1089,93 +1089,95 @@ void cnetlink::port_status_changed(uint32_t port_no,
...
+    return 0;
 
+  std::tuple<uint32_t, enum nbi::port_status, int> change;
   std::deque<std::tuple<uint32_t, enum nbi::port_status, int>> _pc_changes;
-  std::deque<std::tuple<uint32_t, enum nbi::port_status, int>> _pc_retry;
-
-  VLOG(2) << __FUNCTION__;
+  int size = 0;
```
DFritzsche: not needed to declare here

src/netlink/tap_manager.hpp
```diff
@@ -9,8 +9,10 @@
 #include <map>
 #include <memory>
 #include <mutex>
+#include <unordered_map>
 
 #include "sai.hpp"
+#include <netlink/route/link.h>
```
toanju: not needed (even when separated out you should just add a forward declaration for the rtnl_link struct since just the pointer is used here)

src/netlink/tap_manager.hpp
```diff
@@ -67,24 +69,24 @@ class tap_manager final {
...
 private:
   tap_manager(const tap_manager &other) = delete; // non construction-copyable
   tap_manager &operator=(const tap_manager &) = delete; // non copyable
 
   // locked access
   mutable std::mutex tn_mutex; // tap names mutex
-  std::map<std::string, uint32_t> tap_names2id;
-  std::map<std::string, int> tap_names2fds;
+  std::unordered_map<std::string, uint32_t> tap_names2id;
+  std::unordered_map<std::string, int> tap_names2fds;
```
toanju: what is the reason to move to a unordered_map for these?

src/netlink/tap_manager.hpp
```diff
@@ -67,24 +69,24 @@ class tap_manager final {
...
 
   // only accessible from southbound
   std::map<uint32_t, ctapdev *> tap_devs; // port id:tap_device
   std::deque<uint32_t> port_deleted;
 
   // only accessible from cnetlink
-  std::map<int, uint32_t> ifindex_to_id;
-  std::map<uint32_t, int> id_to_ifindex;
+  std::unordered_map<int, uint32_t> ifindex_to_id;
+  std::unordered_map<uint32_t, int> id_to_ifindex;
```
toanju: same question here

PR #172, opened 2019-01-14
l3: added notifications for routes and neighs, minor updates

toanju:
* callbacks can now be registered at l3 to get notified about
routes/neighs.
* get_neighbours_of_route refactored
* minor updates to logging and comments

### DIFF
--- a/src/netlink/nl_l3.cpp
+++ b/src/netlink/nl_l3.cpp
@@ -20,7 +20,6 @@
 #include "nl_output.hpp"
 #include "nl_vlan.hpp"
 #include "sai.hpp"
-#include "tap_manager.hpp"
 #include "utils/rofl-utils.hpp"
 
 namespace std {
@@ -126,7 +125,7 @@ int nl_l3::add_l3_addr(struct rtnl_addr *a) {
   }
 
   bool is_loopback = (rtnl_link_get_flags(link) & IFF_LOOPBACK);
-  bool is_bridge = rtnl_link_is_bridge(link);
+  bool is_bridge = rtnl_link_is_bridge(link); // XXX TODO svi as well?
   int ifindex = 0;
   uint16_t vid = vlan->get_vid(link);
 
@@ -150,7 +149,7 @@ int nl_l3::add_l3_addr(struct rtnl_addr *a) {
 
     if (port_id == 0) {
       if (is_bridge or nl->is_bridge_interface(link)) {
-        LOG(INFO) << __FUNCTION__ << ": host on top of bridge";
+        LOG(INFO) << __FUNCTION__ << ": address on top of bridge";
         port_id = 0;
       } else {
         LOG(ERROR) << __FUNCTION__ << ": invalid port_id 0 for link "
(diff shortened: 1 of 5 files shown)


src/netlink/nl_l3.cpp
```diff
@@ -1075,50 +1105,85 @@ int nl_l3::del_l3_unicast_route(rtnl_route *r) {
...
+    if (nh_addr) {
+      switch (nl_addr_get_family(nh_addr)) {
+      case AF_INET:
+      case AF_INET6:
+        VLOG(2) << __FUNCTION__ << ": ifindex=" << ifindex << " gw=" << nh_addr;
+        break;
+      default:
+        LOG(ERROR) << "gw " << nh_addr
+                   << " unsupported family=" << nl_addr_get_family(nh_addr);
+        nh_addr = nullptr;
```
DFritzsche: this will use nullptr below e.g. logging nullptr, is there flow control missing?

src/netlink/nl_l3.hpp
```diff
@@ -48,16 +49,17 @@ class nl_l3 {
   int add_l3_route(struct rtnl_route *r);
   int del_l3_route(struct rtnl_route *r);
 
-  struct nh_lookup_params {
-    std::deque<struct rtnl_neigh *> *neighs;
-    rtnl_route *rt;
-    cnetlink *nl;
-  };
+  void get_nexthops_of_route(rtnl_route *route,
```
DFritzsche: in most other functions, it is `rtnl_route *r`, could be unified

src/netlink/nl_l3_interfaces.hpp
```diff
@@ -0,0 +1,67 @@
...
+    VLOG(1) << __FUNCTION__ << ": this=" << this;
+    nl_addr_get(nh);
+  }
+
+  nh_stub(const nh_stub &r) : nh(r.nh), ifindex(r.ifindex) {
+    VLOG(1) << __FUNCTION__ << ": this=" << this
+            << ", refcnt=" << nl_object_get_refcnt(OBJ_CAST(nh));
+    nl_addr_get(nh);
+  }
+  // nh_stub(nh_stub &&r) = default;
```
DFritzsche: wild comment?

src/netlink/nl_l3.hpp
DFritzsche: also, this is a get function without return value

src/netlink/nl_l3.cpp
toanju: The ostream operator takes care of this https://github.com/bisdn/basebox/blob/master/src/netlink/nl_output.cpp#L22

src/netlink/nl_l3.hpp
toanju: the second parameter passing the queue that will be filled, otherwise we could move the queue out

src/netlink/nl_l3_interfaces.hpp
toanju: wild thing indeed... played with the move constructor...

src/netlink/nl_l3.cpp
```diff
@@ -1075,50 +1105,85 @@ int nl_l3::del_l3_unicast_route(rtnl_route *r) {
...
-          // lookup neigh in neigh cache, direct?
-          nl_addr *dst = rtnl_route_get_dst(data->rt);
-          if (dst != nullptr) {
-            neigh = data->nl->get_neighbour(ifindex, dst);
-          } else {
-            neigh = nullptr;
-          }
-        }
+    if (pport_id == 0) {
+      VLOG(1) << ": ignoring next hop " << nh;
```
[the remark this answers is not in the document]
toanju: used to be a lambda hence I forgot to add thsi

PR #174, opened 2019-01-16
Correct reinitialization of ports after ofagent restart

rubensfig:


src/netlink/tap_manager.hpp
```diff
@@ -63,6 +63,12 @@ class tap_manager final {
     }
   }
 
+  void empty() noexcept {
```
DFritzsche: I would prefer naming the function `clear`

PR #175, opened 2019-01-22
cnetlink: ignore logical ports status events

toanju:
to reduce some noise


toanju:
sorry this belongs actually to the vxlan PR

PR #177, opened 2019-01-24
add clang-tidy target

toanju:
Run `ninja -C .build clang-tidy` to run clang-tidy on the source files. This depends on clang-tidy and runs faster when [run-clang-tidy.py](https://github.com/llvm-mirror/clang-tools-extra/blob/master/clang-tidy/tool/run-clang-tidy.py) is in your PATH.


rubensfig:
Can we get something on the README stating the clang-tidy optional dependency? That would make it a bit more clear.

PR #181, opened 2019-02-11
Revert "fix: reduce repeated code; ofagent restart no longer causes F…

rubensfig:
…ATAL clause"

This reverts commit 20d47ac4c2888d55f6ad3544f428e024e49aeb1a.


src/of-dpa/controller.cc
```diff
@@ -28,7 +28,7 @@ void controller::handle_conn_established(rofl::crofdpt &dpt,
 void controller::handle_dpt_open(rofl::crofdpt &dpt) {
   std::lock_guard<std::mutex> lock(conn_mutex);
 
-  // Avoid reopening the connection
+   // Avoid reopening the connection
```
toanju:
```suggestion
  // Avoid reopening the connection
```

PR #184, opened 2019-02-15
debugging: update log messages

toanju:


toanju:
neat draft feature... :+1:

PR #201, opened 2019-03-19
bugfix: link local address was additionally adding unrequired VLAN entries

rubensfig:


src/netlink/nl_l3.cc
```diff
@@ -279,6 +279,9 @@ int nl_l3::add_l3_addr_v6(struct rtnl_addr *a) {
                  << ", mask=" << mask;
       return rv;
     }
+
+    // adding link local addresses end here, no further handling required
```
rubensfig:
```suggestion
    // adding link local addresses ends here, no further handling required
```

PR #202, opened 2019-03-19
Vlan refcount

rubensfig:
Counting refcounts for VLANs allows us to remove the L3 address safely, without removing the VLAN entry

### DIFF
--- a/src/netlink/nl_vlan.cc
+++ b/src/netlink/nl_vlan.cc
@@ -17,7 +17,7 @@ namespace basebox {
 
 nl_vlan::nl_vlan(cnetlink *nl) : swi(nullptr), nl(nl) {}
 
-int nl_vlan::add_vlan(rtnl_link *link, uint16_t vid, bool tagged) const {
+int nl_vlan::add_vlan(rtnl_link *link, uint16_t vid, bool tagged) {
   assert(swi);
 
   VLOG(2) << __FUNCTION__ << ": add vid=" << vid << " tagged=" << tagged;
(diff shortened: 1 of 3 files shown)


src/netlink/nl_vlan.h
```diff
@@ -36,6 +37,9 @@ class nl_vlan {
   static const uint16_t vid_high = 0xfff;
   static const uint16_t default_vid = vid_low;
 
+  // ifindex - vlan - refcount
+  std::map<std::pair<uint32_t, uint16_t>, uint32_t> vlan_port;
```
DFritzsche: seeing the types in the map, the name `port_vlan` fits better

src/netlink/nl_vlan.cc
```diff
@@ -76,6 +81,15 @@ int nl_vlan::remove_vlan(rtnl_link *link, uint16_t vid, bool tagged) const {
     return -EINVAL;
   }
 
+  // check for refcount
+  auto key = std::make_pair(port_id, vid);
+  auto refcount = vlan_port.find(key);
+  if (refcount != vlan_port.end()) {
+    refcount->second--;
```
DFritzsche: you should check if 'refcount->second == 1' before, then erase

src/netlink/nl_vlan.cc
```diff
@@ -76,6 +81,15 @@ int nl_vlan::remove_vlan(rtnl_link *link, uint16_t vid, bool tagged) const {
...
+    return rv;
+  } else if (key.second == 1)
```
DFritzsche: this 'else if' hits when 'refcount == vlan_port.end()', as this should not happen printing an error is helpful

src/netlink/nl_vlan.cc
```diff
@@ -76,6 +81,15 @@ int nl_vlan::remove_vlan(rtnl_link *link, uint16_t vid, bool tagged) const {
     return -EINVAL;
   }
 
+  // check for refcount
+  auto key = std::make_pair(port_id, vid);
+  auto refcount = vlan_port.find(key);
+  if (refcount != vlan_port.end()) {
+    refcount->second--;
+    return rv;
```
DFritzsche: rv is assigned 0, why return 0 here (and not proceed)?

also, since you generate rv here (and not passing from another function call) I suggest to make this more clear, like 'return 0' as any (unaware) changes to rv above will change the logic here

PR #213, opened 2019-07-29
WIP: Vxlan examples 

rubensfig:


jklare:
lgtm, I think we need a more detailed description about the setup that is build here, maybe even some graphics that ease the understanding of how and why this is working

PR #230, opened 2019-10-30
QnQ Vlan support

rubensfig:
## Description
This PR adds support to QnQ aware switching. This feature will allow baseboxd to write to the correct OFDPA table, thus setting the correct egress TPID.

The changes in the controller class add a new pair of functions: set_egress_tpid and delete_egress_tpid. These functions have been added to rofl-ofdpa in https://github.com/bisdn/rofl-ofdpa/pull/102, and return the correct flow mod to the switch.

On the netlink part, when a bridge is added with the command `ip l a bridge (...) vlan_protocol 802.1ad`, and ports are subsequently added to it, the sysfs value for the vlan_protocol is read (stored in  /sys/class/net/ + bridge_portname + /bridge/vlan_protocol) to determine if the entry in the TPID table must be set. In the case of a 802.1Q bridge, nothing is set; on a 802.1AD. the TPID entry on the switch is enabled for the attached port.  

When the ports are removed from the bridge, the corresponding entry on the OFDPA tables are removed. When the bridge is removed, all of the ports configured on the bridge will have their entries removed from the tables.  

## Motivation and Context
QinQ VLANs, or 802.1ad is an extension to the VLAN standard that allows multiple VLAN tags to be attached to a single frame. Using stacked VLANs, providers are able to bundle traffic tagged with different VLAN into a single Service tag.

## How Has This Been Tested?
Testing has been based on the added scripts. By configuring the switch using the examples/bridging/qnq/qnq_vlan_controller script, and a server using the examples/bridging/qnq/qnq_vlan_server script, we are able to see that the entries are on the switch, and traffic is flowing as expected.

### DIFF
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -1049,6 +1049,7 @@ void cnetlink::link_deleted(rtnl_link *link) noexcept {
     if (bridge && bridge->is_bridge_interface(link)) {
       LOG(INFO) << __FUNCTION__ << ": deleting bridge";
       vxlan->register_bridge(nullptr);
+      bridge->clear_tpid_entries(); // clear the Egress TPID table
       delete bridge;
       bridge = nullptr;
     }
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -64,20 +64,17 @@ cnetlink::~cnetlink() {
   nl_socket_free(sock_tx);
 }
 
-int cnetlink::load_from_file(const std::string &path) {
-  std::ifstream file;
-  int out = -1;
-  file.open(path);
+int cnetlink::load_from_file(const std::string &path, int base) {
+  std::ifstream file(path);
+  std::string out;
+
   if (file.is_open()) {
     while (!file.eof()) {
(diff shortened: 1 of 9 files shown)


src/sai.h
```diff
@@ -1,5 +1,5 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public * License, v. 2.0.
- * If a copy of the MPL was not distributed with this
+/* This Source Code Form is subject to the terms of the Mozilla Public *
```
KanjiMonster: You now have a stray \* at the end of the line :P

src/netlink/cnetlink.cc
```diff
@@ -1052,6 +1049,7 @@ void cnetlink::link_deleted(rtnl_link *link) noexcept {
     if (bridge && bridge->is_bridge_interface(link)) {
       LOG(INFO) << __FUNCTION__ << ": deleting bridge";
       vxlan->register_bridge(nullptr);
+      bridge->clear_tpid_entries(); // clear the Egress TPID table
```
DFritzsche: isn't this only required for 8021AD bridge case?

PR #231, opened 2019-11-13
Prevent VLAN unaware bridges from being configured

rubensfig:
## Description
When a user tries to configure traditional bridges, and attach baseboxd's ports to it, the switch will accept it, but leads to errors, since this configuration is not currently supported in baseboxd. Therefore, reading the vlan_filtering option in the bridge interface sysfs is necessary to prevent misconfiguration. 

This PR also adds further exception handling to the functions introduced in #230,  set_egress_tpid, and delete_ingress_pid.

## Motivation and Context

This PR aims to introduce a feature where baseboxd will prevent the user from configuring non-vlan aware bridges, since these can be created and used, but will lead to failures if that happens. 

## How Has This Been Tested?
The following list of commands should work correctly, resuming normal baseboxd's operation after.

```
ip link add swbridge type bridge vlan_filtering 1
ip link set port1 master swbridge
ip link delete swbridge
```

The following list of commands should now not work, causing baseboxd to log fatal the message: 
`unsupported: bridge configured with vlan_filtering 0`.

```
ip link add swbridge type bridge 
ip link set port1 master swbridge
ip link delete swbridge
```

### DIFF
--- a/src/netlink/nl_bridge.cc
+++ b/src/netlink/nl_bridge.cc
@@ -42,6 +42,10 @@ void nl_bridge::set_bridge_interface(rtnl_link *bridge) {
 
   this->bridge = bridge;
   nl_object_get(OBJ_CAST(bridge));
+
+  if (!get_vlan_filtering())
+    LOG(FATAL) << __FUNCTION__
+               << " unsupported: bridge configured with vlan_filtering 0";
 }
 
 bool nl_bridge::is_bridge_interface(rtnl_link *link) {


src/of-dpa/controller.cc
```diff
@@ -1728,6 +1728,12 @@ int controller::set_egress_tpid(uint32_t port) noexcept {
     dpt.send_flow_mod_message(rofl::cauxid(0),
                               fm_driver.set_port_tpid(dpt.get_version(), port));
 
+  } catch (rofl::eRofBaseNotFound &e) {
+    LOG(ERROR) << ": caught rofl::eRofBaseNotFound";
+    rv = -EINVAL;
+  } catch (rofl::eRofConnNotConnected &e) {
+    LOG(ERROR) << ": not connected msg=" << e.what();
+    rv = -ENOTCONN;
   } catch (rofl::eRofBaseNotFound &e) {
```
KanjiMonster: I not very versed in C++ exception handling, but isn't `rofl::eRofBaseNotFound &e` already caught here? Same for the second change.

KanjiMonster:
LGTM

PR #235, opened 2019-12-09
meson: fix rofl-ofdpa dependency

rubensfig:
## Description
Fixed rofl-ofdpa dependency to reflect new version. Added meson check for version.

## Motivation and Context
There was a costumer request claiming that basebox was breaking on executing a specific rofl-ofdpa call, so maintaining the version dependency will ensure correct version.

## How Has This Been Tested?
N/A


jklare:
lgtm

Issue #234, opened 2019-11-22
PVID Egress untagged bug using systemd-networkd


DFritzsche:
## Expected Behavior
Seeing output of "bridge vlan":
```
swbridge	 1 PVID Egress Untagged
	 15
	 4090

port1	 1 Egress Untagged
	 15 PVID Egress Untagged
```

you would expect `newVlanId = 0x100f (VLAN 15)` set in the switche's flowtable
```
--  inPort = 1 (Physical)  vlanId:mask = 0x0000:0x1fff (VLAN 0) | GoTo = 20 (Termination MAC) newVlanId = 0x100f (VLAN 15) | priority = 3 hard_time = 0 idle_time = 0 cookie = 10
```

## Actual Behavior

in some cases it does not set `newVlanId = 0x100f (VLAN 15)` but `newVlanId = 0x1001 (VLAN 1)` instead

```
--  inPort = 1 (Physical)  vlanId:mask = 0x0000:0x1fff (VLAN 0) | GoTo = 20 (Termination MAC) newVlanId = 0x1001 (VLAN 1) | priority = 3 hard_time = 0 idle_time = 0 cookie = 19
```

## Steps to Reproduce the Problem

copy the attached systemd-networkd configuration files to /etc/systemd/network and reboot the switch
[network.zip](https://github.com/bisdn/basebox/files/3880488/network.zip)

## Specifications
  - Version:
baseboxd version 1.1.1
  - Platform: 
ID="bisdn-linux"
NAME="BISDN Linux"
VERSION="19.06 (Marinara)"
VERSION_ID="19.06"
PRETTY_NAME="BISDN Linux 19.06 (Marinara)"
BUILD_ID="20191121222227"

  - Subsystem:

DFritzsche:
Solved in: https://github.com/bisdn/basebox/pull/241

PR #236, opened 2020-01-20
nl_l3: do not create termination after registering link local addresses;

rubensfig:
## Description
Adding a termination MAC entry after registering the link local address has a bug, where the wrong VID is discovered therefore writing the incorrect VLAN on the TMAC entry on the ASIC. 

This PR fixes VID discovery for the bridged ports, thus being able to write the correct VID. Also featured on this PR is a simplification of the addition of IP addresses, IPv4 and IPv6, and  

## Motivation and Context
See #234. 
It effects not only systemd-networkd but also iproute2.

Configuring access ports on the switch means adding the PVID on the following rule,

inPort = ABC (Physical)  vlanId:mask = 0x0000:0x1fff (VLAN 0) | GoTo = 20 (Termination MAC) newVlanId = 0x1064 (VLAN XXX) | priority = 3 hard_time = 0 idle_time = 0 cookie = 21,

where XXX is then the PVID for a specific port ABC.

This bug i that this rule was written correctly, but then overwritten by the re-addition of IP addresses, like the IPv6 link local. Since baseboxd also configures the VLAN entries on this case, it was re-writing the correct rule with the wrongly discovered VLAN id.

## How Has This Been Tested?
PR includes a script under bridging/trunk/cntl_trunk_script.sh

### DIFF
--- a/examples/bridging/trunk/cntl_trunk_script.sh
+++ b/examples/bridging/trunk/cntl_trunk_script.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+PORTA=${PORTA:-port31}
+PORTB=${PORTB:-port32}
+PORTC=${PORTC:-port33}
+PORTD=${PORTD:-port34}
+BRIDGE=${BRIDGE:-swbridge}
+
+function setup() {
+  ip link add name ${BRIDGE} type bridge vlan_filtering 1
+  ip link set ${BRIDGE} up
+
+  # port 1
+  ip link set ${PORTA} master ${BRIDGE}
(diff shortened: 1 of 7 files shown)


examples/bridging/trunk/cntl_trunk_script.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
```
KanjiMonster: Maybe add a (short) description what this example script does and what the expected setup is? Most other examples seem to have that

PR #244, opened 2020-02-21
tap_man: avoid changing interface state on modify

rubensfig:
## Description
This PR removes the modify port state capability from baseboxd. Due to not being able to accurately represent link admin and phy state, changing the port status on every notification will ensure that port configuration is removed, which brings the switch to a weird state on the retriggering of  the port state.

## Motivation and Context
Rebooting servers must not change port state on the switch, to avoid losing configuration.

## How Has This Been Tested?
Tested against BISDN Linux v2.1.0

```
10-swbridge.netdev
[NetDev]
Name=swbridge
Kind=bridge

[Bridge]
DefaultPVID=0
VLANFiltering=1
```
```
10-swbridge.network
[Match]
Name=swbridge

[BridgeVLAN]
VLAN=108
````
```
20-port1.network
[Match]
Name=port1

[Network]
Bridge=swbridge

[BridgeVLAN]
VLAN=108
```
```
20-port3.network
[Match]
Name=port3

[Network]
Bridge=swbridge

[BridgeVLAN]
PVID=108
EgressUntagged=108

```

### DIFF
--- a/src/netlink/nbi_impl.cc
+++ b/src/netlink/nbi_impl.cc
@@ -52,7 +52,7 @@ void nbi_impl::port_notification(
     case PORT_EVENT_MODIFY:
       switch (get_port_type(ntfy.port_id)) {
       case nbi::port_type_physical:
-        tap_man->change_port_status(ntfy.name, ntfy.status);
+        LOG(INFO) << __FUNCTION__ << ": port state changed, continuing";
         break;
       default:
         LOG(ERROR) << __FUNCTION__ << ": unknown port";
--- a/src/netlink/nbi_impl.cc
+++ b/src/netlink/nbi_impl.cc
@@ -58,10 +58,14 @@ void nbi_impl::port_notification(
         LOG(ERROR) << __FUNCTION__ << ": unknown port";
         break;
       }
+      break;
     case PORT_EVENT_ADD:
       switch (get_port_type(ntfy.port_id)) {
       case nbi::port_type_physical:
-        tap_man->create_tapdev(ntfy.port_id, ntfy.name, *this);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/netlink/nbi_impl.cc [hunk 1]
```diff
@@ -49,15 +49,6 @@ void nbi_impl::port_notification(
 
   for (auto &&ntfy : notifications) {
     switch (ntfy.ev) {
-    case PORT_EVENT_MODIFY:
```
DFritzsche: think about removing the PORT_EVENT_MODIFY case from sai.h then

KanjiMonster: Maybe we should keep the PORT_EVENT_MODIFY case but just have a comment that we cannot reflect the link state for tap devices?

src/netlink/nbi_impl.cc
```diff
@@ -49,15 +49,6 @@ void nbi_impl::port_notification(
...
-      case nbi::port_type_physical:
-        tap_man->change_port_status(ntfy.name, ntfy.status);
-        break;
-      default:
-        LOG(ERROR) << __FUNCTION__ << ": unknown port";
-        break;
-      }
     case PORT_EVENT_ADD:
       switch (get_port_type(ntfy.port_id)) {
       case nbi::port_type_physical:
```
KanjiMonster: Also `tap_man->change_port_status(ntfy.name, ntfy.status);` is still called here (below this line) - is this okay?

rubensfig: This one is used for setting the initial state on the interfaces after creation

PR #248, opened 2020-03-24
baseboxd: add Multicast Support on bridges

rubensfig:
# Description

In order to develop multicast, baseboxd must be aware of the modifications in the Linux kernel bridge driver multicast database. 

This database tracks the multicast memberships of ports in a certain bridge. The following is the available data in this database. 

```
dev swbridge 
port port7 
grp 239.1.0.31 
state temp
vid 108
```

This database is filled either manually via 

```
bridge mdb add dev <bridge> port <port> vid <VLAN id> grp <multicast group>
```

or via the incoming IGMP/MLD notifications from the hosts. These protocols allow the router to know which multicast group a certain host is interested in joining. 

The Linux kernel has a notification system via Netlink that reports the changes in the multicast database over the [RTNLGRP_MDB](https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/rtnetlink.h#L711) multicast socket.

As such notifications such as RTM_NEWMDB, RTM_GETMDB can be processed in userspace and in baseboxd specifically. 

libn has a message caching mechanism that baseboxd is employing for Link, Address, Route and Neighbor notifications. The MDB cache is the new addition to both libnl and baseboxd that must be implemented. Refer to [libnl git PR](https://github.com/thom311/libnl/pull/234).

baseboxd is then able to obtain the relevant information from the Linux kernel, and translate it to the ASIC via OpenFlow.

OFDPA uses the following model: [rofl-ofdpa PR](https://github.com/bisdn/rofl-ofdpa/pull/104)

```
                                          / if match Unicast MAC   -> Unicast Routing (30)
Ingress VLAN (10) -> Termination MAC (20) -> if match Multicast MAC -> Multicast Routing (40)
                                          \ Default miss -> Bridging (50)
```

After matching on the Multicast Routing table there is an optional Openflow action to direct to a L3 Multicast Group. 
IP multicast packets are handled differently if they are switched (same VLAN) or routed(different VLAN). 
- Switched: cannot be output to IN_PORT (ingress port)
- Routed: must be allowed to egress via IN_PORT (ingress port)

OFDPA handles this split by chaining the L3 Multicast group to the respective L2 or L3 interface group entry. This decision where to the correctly forward the multicast driver is then based on the VLAN.

### DIFF
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -15,9 +15,10 @@
 #include <netlink/object.h>
 #include <netlink/route/addr.h>
 #include <netlink/route/link.h>
-#include <netlink/route/link/vlan.h>
 #include <netlink/route/link/bonding.h>
+#include <netlink/route/link/vlan.h>
 #include <netlink/route/link/vxlan.h>
+#include <netlink/route/mdb.h>
 #include <netlink/route/neighbour.h>
 #include <netlink/route/route.h>
 #include <systemd/sd-daemon.h>
@@ -156,6 +157,20 @@ void cnetlink::init_caches() {
                                   (change_func_v2_t)&nl_cb_v2, this);
   if (0 != rc) {
     LOG(FATAL) << __FUNCTION__ << ": add route/neigh to cache mngr";
+  }
+
+  /* init mdb cache */
+  rc = rtnl_mdb_alloc_cache_flags(sock_mon, &caches[NL_MDB_CACHE],
+                                  NL_CACHE_AF_ITER);
+  if (0 != rc) {
+    LOG(FATAL) << __FUNCTION__
+               << ": rtnl_mdb_alloc_cache_flags failed rc=" << rc;
+  }
+  rc = nl_cache_mngr_add_cache_v2(mngr, caches[NL_MDB_CACHE],
+                                  (change_func_v2_t)&nl_cb_v2, this);
+  if (0 != rc) {
+    LOG(FATAL) << __FUNCTION__
+               << ": nl_cache_mngr_add_cache_v2: add route/mdb to cache mngr";
   }
 
   try {
@@ -425,6 +440,10 @@ void cnetlink::handle_wakeup(rofl::cthread &thread) {
     case RTM_NEWADDR:
     case RTM_DELADDR:
       route_addr_apply(obj);
+      break;
+    case RTM_NEWMDB:
+    case RTM_DELMDB:
+      route_mdb_apply(obj);
       break;
     default:
       LOG(ERROR) << __FUNCTION__ << ": unexpected netlink type "
@@ -1245,6 +1264,29 @@ std::deque<rtnl_neigh *> cnetlink::search_fdb(uint16_t vid, nl_addr *lladdr) {
   }
 
   return fdb_entries;
+}
+
+void cnetlink::route_mdb_apply(const nl_obj &obj) {
+
+  switch (obj.get_action()) {
+  case NL_ACT_NEW:
+    assert(obj.get_new_obj());
+    LOG(INFO) << __FUNCTION__ << ": new mdb entry";
+
+    if (bridge)
+      bridge->mdb_entry_add(MDB_CAST(obj.get_new_obj()));
+    break;
+  case NL_ACT_DEL:
+    assert(obj.get_old_obj());
+    LOG(INFO) << __FUNCTION__ << ": deleting mdb entry";
+
+    if (bridge)
+      bridge->mdb_entry_remove(MDB_CAST(obj.get_old_obj()));
+    break;
+  default:
+    LOG(ERROR) << __FUNCTION__ << ": invalid action " << obj.get_action();
+    break;
+  }
 }
 
 } // namespace basebox
--- a/src/netlink/netlink-utils.cc
+++ b/src/netlink/netlink-utils.cc
@@ -10,6 +10,7 @@
 #include <cstring>
 
 #include <glog/logging.h>
+#include <linux/if_ether.h>
 #include <netlink/route/link.h>
 
 #include "netlink-utils.h"
@@ -108,6 +109,27 @@ uint64_t nlall2uint64(const nl_addr *a) noexcept {
   b_[1] = a_[4];
   b_[0] = a_[5];
   return b;
+}
+
+// Convert Multicast IPv4 addresses to MAC address
+// Multicast MAC address : 01:00:53:XX:XX:XX
+void multicast_ipv4_to_ll(const uint32_t addr, unsigned char *dst) noexcept {
+  dst[0] = 0x01;
+  dst[1] = 0x00;
+  dst[2] = 0x5e;
+  dst[3] = (addr >> 16) & 0x7f;
+  dst[4] = (addr >> 8) & 0xff;
+  dst[5] = addr & 0xff;
+}
+
+// Convert Multicast IPv6 addresses to MAC address
+// Multicast MAC address : 33:33:XX:XX:XX:XX
+void multicast_ipv6_to_ll(const struct in6_addr *addr,
+                          unsigned char *dst) noexcept {
+  dst[0] = 0x33;
+  dst[1] = 0x33;
+
+  memcpy(dst + 2, &addr->s6_addr32[3], sizeof(__u32));
 }
 
 } // namespace basebox
--- a/src/netlink/nl_bridge.cc
+++ b/src/netlink/nl_bridge.cc
@@ -6,9 +6,11 @@
 #include <cstring>
 #include <map>
 #include <utility>
+#include <linux/if_packet.h>
 
 #include <glog/logging.h>
 #include <netlink/route/link.h>
+#include <netlink/route/mdb.h>
 #include <netlink/route/link/vlan.h>
 #include <netlink/route/link/vxlan.h>
 #include <netlink/route/neighbour.h>
@@ -35,6 +37,15 @@ nl_bridge::nl_bridge(switch_interface *sw, std::shared_ptr<tap_manager> tap_man,
 }
 
 nl_bridge::~nl_bridge() { rtnl_link_put(bridge); }
+
+static rofl::caddress_in4 libnl_in4addr_2_rofl(struct nl_addr *addr, int *rv) {
+  struct sockaddr_in sin;
+  socklen_t salen = sizeof(sin);
+
+  assert(rv);
+  *rv = nl_addr_fill_sockaddr(addr, (struct sockaddr *)&sin, &salen);
+  return rofl::caddress_in4(&sin, salen);
+}
 
 void nl_bridge::set_bridge_interface(rtnl_link *bridge) {
   assert(bridge);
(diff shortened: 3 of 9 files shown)


DFritzsche:
Link to OF-DPA MR points to libnl.

VERSION
```diff
@@ -1 +1 @@
-1.4.1
+1.5.0
```
DFritzsche, line 1: please also update the dependencies (for rofl-ofdpa and libnl)

src/netlink/nl_bridge.cc
```diff
@@ -770,4 +785,136 @@ void nl_bridge::clear_tpid_entries() {
...
+      [](struct rtnl_mdb_entry *it, void *arg) {
+        auto m_database =
+            static_cast<std::deque<struct rtnl_mdb_entry *> *>(arg);
+        m_database->push_back(it);
+      },
+      &_mdb);
+
+  for (auto i : _mdb) {
+    int _port_ifindex = rtnl_mdb_entry_get_ifindex(i);
+    int _port_id = nl->get_port_id(_port_ifindex);
```
DFritzsche, line 813: _port_id is of type uint32_t when used above. Please also check the type of the other variables, _vid probably should be uint16_t

src/netlink/nl_bridge.cc
```diff
@@ -770,4 +785,136 @@ void nl_bridge::clear_tpid_entries() {
...
+    int _vid = rtnl_mdb_entry_get_vid(i);
```
DFritzsche, line 868: types again

src/netlink/nl_bridge.cc
```diff
@@ -770,4 +785,136 @@ void nl_bridge::clear_tpid_entries() {
...
+
+    auto _addr = rtnl_mdb_entry_get_addr(i);
+    if (rtnl_mdb_entry_get_proto(i) == ETH_P_IP) {
+      rofl::caddress_in4 ipv4_dst = libnl_in4addr_2_rofl(_addr, &rv);
+
+      unsigned char buf[ETH_ALEN];
+      multicastv42ll(ipv4_dst.get_addr_hbo(), buf);
```
DFritzsche, line 875: is 2 == "to" again?

src/netlink/nl_bridge.cc
```diff
@@ -770,4 +785,136 @@ void nl_bridge::clear_tpid_entries() {
...
+      if (rv < 0) {
+        LOG(ERROR) << __FUNCTION__ << ": failed to add bridging MC entry";
+        return rv;
+      }
+
+      rv = sw->l2_multicast_group_remove(_vid, _port_id, ipv4_dst);
+
+      // if the group still has entries do not remove the bridging entry
+      if (rv == -ENOTEMPTY) {
+        LOG(WARNING) << __FUNCTION__ << ": group still exists ";
```
DFritzsche, line 888: I don't think we need to handle -ENOTEMPTY if the group still exists, since we don't do this in other parrts of baseboxd. It would be sufficient to log when it has been deleted (or created), in all other cases we assume the group is still there. Also, why is it loglevel warning?

Instead I would log if rv == 0 (nothing to remove) because this indicates we made an error elsewhere (we try to remove something that does not exists)

src/netlink/nl_bridge.cc
```diff
@@ -770,4 +785,136 @@ void nl_bridge::clear_tpid_entries() {
...
+
+int nl_bridge::process_mdb_entry(rtnl_mdb *mdb_entry) {
+  LOG(INFO) << __FUNCTION__ << ": adding mdb entry";
+
+  int rv = 0;
+  int _bridge = rtnl_mdb_get_ifindex(mdb_entry);
+
+  if (!is_bridge_interface(_bridge)) {
+    LOG(ERROR) << ": bridge not set correctly";
+    return rv;
```
DFritzsche, line 914: So this function always returns 0?

src/netlink/nl_bridge.cc
```diff
@@ -59,6 +64,16 @@ bool nl_bridge::is_bridge_interface(rtnl_link *link) {
   return true;
 }
 
+bool nl_bridge::is_bridge_interface(const int ifindex) {
+  assert(link);
```
DFritzsche, line 68: what link does this assert evaluate?

src/netlink/nl_bridge.h
```diff
@@ -43,6 +44,7 @@ class nl_bridge {
 
   void set_bridge_interface(rtnl_link *);
   bool is_bridge_interface(rtnl_link *);
+  bool is_bridge_interface(const int);
```
DFritzsche, line 47: this function is not used, and wouldn't it be better to use the one above? bool is_bridge_interface(rtnl_link *);

src/netlink/nl_l3.cc
```diff
@@ -986,11 +986,13 @@ int nl_l3::add_l3_route(struct rtnl_route *r) {
   case RTN_UNICAST:
     rv = add_l3_unicast_route(r, false);
     break;
+  case RTN_MULTICAST:
+    VLOG(1) << " ADDED MULTICAST ROUTE ";
```
DFritzsche, line 990: should we have a message for the delete as well?

src/netlink/cnetlink.cc
```diff
@@ -149,14 +150,27 @@ void cnetlink::init_caches() {
                                     NL_CACHE_AF_ITER);
   if (0 != rc) {
     LOG(FATAL) << __FUNCTION__
-               << ": rtnl_link_alloc_cache_flags failed rc=" << rc;
+               << ": rtnl_neigh_alloc_cache_flags failed rc=" << rc;
```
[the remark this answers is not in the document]
rubensfig: Its logs the correct function that failed.

src/netlink/nl_bridge.cc
```diff
@@ -36,6 +38,9 @@ nl_bridge::nl_bridge(switch_interface *sw, std::shared_ptr<tap_manager> tap_man,
 
 nl_bridge::~nl_bridge() { rtnl_link_put(bridge); }
 
+// Copy to netlink-utils.h
+static rofl::caddress_in4 libnl_in4addr_2_rofl(struct nl_addr *addr, int *rv);
```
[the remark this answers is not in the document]
rubensfig: historical reasons, see https://github.com/bisdn/basebox/blob/master/src/netlink/nl_l3.cc#L87

src/netlink/netlink-utils.cc
```diff
@@ -110,4 +111,19 @@ uint64_t nlall2uint64(const nl_addr *a) noexcept {
   return b;
 }
 
+// Convert Multicast IP v4 addresses to MAC address
+// Multicast MAC address : 01:00:53:XX:XX:XX
+int multicastv42ll(uint32_t addr, unsigned char *mc_addr) noexcept {
+  mc_addr[0] = 0x01;
+  mc_addr[1] = 0x00;
+  mc_addr[2] = 0x5e;
+  mc_addr[5] = addr & 0xFF;
```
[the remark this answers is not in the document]
rubensfig: The order is like this because of the MAC address order. The three first bytes of the Multicast MAC are 01:00:5e, so in my opinion makes a bit more sense to maintain this structure.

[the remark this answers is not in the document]
KanjiMonster: I agree with @DFritzsche. I would suggest 
```
  mc_addr[3] = (addr >> 16) & 0x7F;
  mc_addr[4] = (addr >> 8) & 0xFF;
  mc_addr[5] = addr & 0xFF;
```
which would be less code, more readable IMHO, and you could make addr `const`, since you don't manipulate it anymore.

Also maybe decide if you capitalize hex numbers or not.

src/netlink/netlink-utils.cc
```diff
@@ -110,4 +111,19 @@ uint64_t nlall2uint64(const nl_addr *a) noexcept {
...
+  addr >>= 8;
+  mc_addr[4] = addr & 0xFF;
+  addr >>= 8;
+  mc_addr[3] = addr & 0x7F;
+
+  return 0;
```
KanjiMonster, line 126: Since this function currently always returns 0, maybe change its return type to void and remove the return statement.

KanjiMonster: Looking at the places where you use it, you even ignore the return value, so make that a strong suggestion.

src/of-dpa/controller.cc
```diff
@@ -762,14 +762,160 @@ int controller::l2_overlay_addr_remove(uint32_t tunnel_id, uint32_t lport_id,
...
+
+    if (dom_it->second.size()) {
+      dpt.send_group_mod_message(
+          rofl::cauxid(0),
+          // send update without port
+          fm_driver.enable_group_l2_multicast(dpt.get_version(), vid, vid,
+                                              dom_it->second, true));
+
+    } else {
+      LOG(WARNING) << __FUNCTION__ << ": deleting group ";
```
KanjiMonster, line 828: This doesn't seem like something to warn about?

src/netlink/netlink-utils.cc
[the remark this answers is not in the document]
DFritzsche: ok, its a copy of this one: https://elixir.bootlin.com/linux/v4.6/source/include/net/ip.h#L392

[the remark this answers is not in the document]
rubensfig: Yeah, it was taken from there. Either way readability is improved

src/of-dpa/controller.cc
```diff
@@ -762,14 +762,159 @@ int controller::l2_overlay_addr_remove(uint32_t tunnel_id, uint32_t lport_id,
...
+          rofl::cauxid(0),
+          // send update without port
+          fm_driver.enable_group_l2_multicast(dpt.get_version(), vid, vid,
+                                              dom_it->second, true));
+
+    } else {
+      dpt.send_group_mod_message(
+          rofl::cauxid(0),
+          // send update without port
+          fm_driver.enable_group_l2_multicast(dpt.get_version(), vid, vid,
```
DFritzsche, line 831: so this is not needed, only the barrier below matters

src/of-dpa/controller.cc
```diff
@@ -762,14 +762,159 @@ int controller::l2_overlay_addr_remove(uint32_t tunnel_id, uint32_t lport_id,
...
+    if (dom_it == mc_groups.end()) {
+      std::set<uint32_t> empty_set;
+      dom_it = mc_groups.emplace(mc_group, empty_set).first;
+    }
+
+    // insert group_id to set
+    dom_it->second.insert(fm_driver.group_id_l2_interface(id, vid));
+
+    // create/update new L2 Multicast group
+    dpt.send_group_mod_message(rofl::cauxid(0),
```
DFritzsche, line 784: suggest to add a barrier request after to ensure the group can be referenced right away, e.g. by the bridging entry

src/netlink/netlink-utils.cc
```diff
@@ -110,4 +111,25 @@ uint64_t nlall2uint64(const nl_addr *a) noexcept {
   return b;
 }
 
+// Convert Multicast IPv4 addresses to MAC address
+// Multicast MAC address : 01:00:53:XX:XX:XX
+void multicast_ipv4_to_ll(const uint32_t addr, unsigned char *dst) noexcept {
```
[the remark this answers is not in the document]
rubensfig: Well, the struct im passing to the IPv6 function is actually the `struct in6_addr`. This is to align with the Linux kernel version of this code.

In this case, i dont see much sense in transforming the object to the `rofl::caddress_in4` struct, because we already have the binary representation of the IP address in both cases.

PR #250, opened 2020-04-15
tap_manager: set port speed on tap interfaces

rubensfig:
## Motivation and Context

Link speed must be adapted to the correct reported speed from OFDPA. Currently the device is stuck at 10MB/s

```
ethtool port1 | grep Speed
        Speed: 10Mb/s
```

Link speed for tap interfaces looks to be set at https://elixir.bootlin.com/linux/latest/source/drivers/net/tun.c#L3519

ethtool is able to change the speed with the ioctl `ioctl(3, SIOCETHTOOL, 0x7ffe47542930)`

The link speed is already reported from the switch, so baseboxd is capable of setting this information in Linux.

ethtool command: `ethtool -s port1 speed=100`

## How Has This Been Tested?
In Basebox, run 

```
# client_drivshell port xe53 speed=10000
# client_drivshell port xe53 speed
PORT: Status (* indicates PHY link up)
 *xe53    (40G) 


# ethtool port54
Settings for port54:
        Supported ports: [ ]
        Supported link modes:   Not reported
        Supported pause frame use: No
        Supports auto-negotiation: No
        Supported FEC modes: Not reported
        Advertised link modes:  Not reported
        Advertised pause frame use: No
        Advertised auto-negotiation: No
        Advertised FEC modes: Not reported
        Speed: 10000Mb/s
        Duplex: Full
        Port: Twisted Pair
        PHYAD: 0
        Transceiver: internal
        Auto-negotiation: off
        MDI-X: Unknown
        Current message level: 0xffffffa1 (-95)
                               drv ifup tx_err tx_queued intr tx_done rx_status pktdata hw wol 0xffff8000
        Link detected: yes
```

### DIFF
--- a/src/netlink/nbi_impl.cc
+++ b/src/netlink/nbi_impl.cc
@@ -38,6 +38,7 @@ void nbi_impl::port_notification(
       switch (get_port_type(ntfy.port_id)) {
       case nbi::port_type_physical:
         tap_man->change_port_status(ntfy.name, ntfy.status);
+	tap_man->set_port_speed(ntfy.name, ntfy.speed, ntfy.duplex);
         break;
       default:
         LOG(ERROR) << __FUNCTION__ << ": unknown port";
@@ -49,6 +50,7 @@ void nbi_impl::port_notification(
       case nbi::port_type_physical:
         tap_man->create_tapdev(ntfy.port_id, ntfy.name, *this);
         tap_man->change_port_status(ntfy.name, ntfy.status);
+        tap_man->set_port_speed(ntfy.name, ntfy.speed, ntfy.duplex);
         break;
       case nbi::port_type_vxlan:
         // XXX TODO notify this?
--- a/src/netlink/tap_manager.cc
+++ b/src/netlink/tap_manager.cc
@@ -8,11 +8,16 @@
 #include <netlink/route/link.h>
 #include <sys/ioctl.h>
 #include <utility>
+#include <linux/ethtool.h>
+#include <linux/sockios.h>
 
 #include "cnetlink.h"
 #include "ctapdev.h"
 #include "tap_io.h"
 #include "tap_manager.h"
+
+#define ETHTOOL_LINK_MODE_MASK_MAX_KERNEL_NU32 (SCHAR_MAX)
+#define ETHTOOL_SPEED(speed) speed / 1000 // conversion to Mbit
 
 namespace basebox {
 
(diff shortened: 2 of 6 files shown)


src/netlink/tap_manager.cc
```diff
@@ -336,4 +339,54 @@ int tap_manager::change_port_status(const std::string name, bool status) {
   return error;
 }
 
+#define ETH_SPEED(speed) speed / 1000 // conversion to Mbyte
```
rr-santos, line 342: Shouldn't this be placed in the first lines of the file?

KanjiMonster: Also I would assume the speed is given in Mbit/s, not MByte/s.

src/netlink/tap_manager.cc
```diff
@@ -336,4 +339,54 @@ int tap_manager::change_port_status(const std::string name, bool status) {
...
+int tap_manager::change_port_speed(const std::string name, uint32_t speed) {
+  struct ifreq ifr = {};
+
+  auto sockFd = socket(AF_INET, SOCK_DGRAM, 0);
+  strncpy(ifr.ifr_name, name.c_str(), IFNAMSIZ);
+
+  // The ethtool API requires an handshake via ETHTOOL_GLINKSETTINGS
```
rr-santos, line 349: Should the ethtool handshake be in a separate function? That'd be useful if somehow we use more ethtool configuration features.

src/netlink/tap_manager.cc
```diff
@@ -336,4 +339,54 @@ int tap_manager::change_port_status(const std::string name, bool status) {
...
+  // before getting the "real" value to agree on the length of link
+  // mode bitmaps.
+
+  struct link_settings {
```
rr-santos, line 353: Could we move this struct to the headers?

src/netlink/tap_manager.cc
```diff
@@ -336,4 +339,54 @@ int tap_manager::change_port_status(const std::string name, bool status) {
...
+    __u32 link_mode_data[3 * ETHTOOL_LINK_MODE_MASK_MAX_KERNEL_NU32];
+  } link_settings;
+
+  memset(&link_settings, 0, sizeof(struct link_settings));
+  link_settings.req.cmd = ETHTOOL_GLINKSETTINGS;
+
+  ifr.ifr_data = reinterpret_cast<char *>(&link_settings);
+  int error = ioctl(sockFd, SIOCETHTOOL, static_cast<void *>(&ifr));
+  if (error < 0) {
+    VLOG(3) << " wrong ioctl, error=" << error;
```
KanjiMonster, line 364: All three error messages have identical text, so you won't be able to tell which ioctl failed. Maybe include the ioctl name in the error message for easier debugging?

src/netlink/tap_manager.cc
```diff
@@ -336,4 +339,54 @@ int tap_manager::change_port_status(const std::string name, bool status) {
...
+  }
+
+  link_settings.req.speed = ETH_SPEED(speed);
+  link_settings.req.cmd = ETHTOOL_SLINKSETTINGS;
+
+  ifr.ifr_data = reinterpret_cast<char *>(&link_settings);
+  error = ioctl(sockFd, SIOCETHTOOL, static_cast<void *>(&ifr));
+  if (error < 0) {
+    VLOG(1) << " wrong ioctl, error=" << error;
```
KanjiMonster, line 386: This one is `VLOG(1)`, while the other two are `VLOG(3)` - intentional?

src/netlink/tap_manager.cc
rubensfig: This was adapted from https://github.com/systemd/systemd/blob/a8cff0034c8271e0042ce6116fd45b5405ae4de3/src/shared/ethtool-util.c#L669. 

Im not sure if we need to export this struct to the class headers, since its only being used here.

src/netlink/tap_manager.cc
rubensfig: No strong opinions on this. I think for now its fine.

src/netlink/nbi_impl.cc
```diff
@@ -52,7 +52,8 @@ void nbi_impl::port_notification(
     case PORT_EVENT_MODIFY:
       switch (get_port_type(ntfy.port_id)) {
       case nbi::port_type_physical:
-        LOG(INFO) << __FUNCTION__ << ": port state changed, continuing";
+        LOG(INFO) << __FUNCTION__ << ": changing port speed";
+        tap_man->change_port_speed(ntfy.name, ntfy.speed);
```
DFritzsche, line 56: So, the function SETs the port speed regardless whether there is a change. Also it doesn't check if there was a change to port speed. The PORT_EVENT_MODIFY event could be triggered by something that is not related to port speed at all. I recommend naming the function `set_port_speed`.

src/of-dpa/controller.cc
```diff
@@ -276,9 +277,11 @@ void controller::handle_port_desc_stats_reply(
 
     bool status = (!(port.get_config() & rofl::openflow13::OFPPC_PORT_DOWN) &&
                    !(port.get_state() & rofl::openflow13::OFPPS_LINK_DOWN));
+    uint32_t speed = port.get_ethernet().get_curr_speed();
```
KanjiMonster, line 280: 1G ports can be also in half duplex mode, which you currently don't check Since there is no extra field for that, you will need to look at `cofportdesc_prop_ethernet::curr`, and check if any of the bits 0, 2 or 4 are set (`OFPPF_10MB_HD`, `OFPPF_100MB_HD` and `OFPPF_1GB_HD`).

src/netlink/tap_manager.cc
```diff
@@ -336,4 +340,54 @@ int tap_manager::change_port_status(const std::string name, bool status) {
...
+    VLOG(3) << __FUNCTION__ << " failed to get port info, error=" << error;
+  }
+
+  link_settings.req.speed = ETHTOOL_SPEED(speed);
+  link_settings.req.cmd = ETHTOOL_SLINKSETTINGS;
+
+  ifr.ifr_data = reinterpret_cast<char *>(&link_settings);
+  error = ioctl(sockFd, SIOCETHTOOL, static_cast<void *>(&ifr));
+  if (error < 0) {
+    LOG(ERROR) << __FUNCTION__ << " failed to set port speed, error=" << error;
```
KanjiMonster, line 387: You should log for which port this failed.

src/netlink/tap_manager.h
```diff
@@ -76,6 +77,7 @@ class tap_manager final {
   int get_fd(uint32_t port_id) const noexcept;
 
   int change_port_status(const std::string name, bool status);
+  int set_port_speed(const std::string name, uint32_t speed);
```
KanjiMonster, line 80: Maybe it would make sense to merge these into one function? E.g.

```
    int change_port_status(const std::string name, bool link, uint32_t speed, bool fullduplex);
```

rubensfig: I think since we are addressing two different ioctls it makes sense to keep it separated.

PR #253, opened 2020-05-22
baseboxd: delay processing of netlink messages

rubensfig:

### DIFF
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -500,10 +500,7 @@ void cnetlink::nl_cb_v2(struct nl_cache *cache, struct nl_object *old_obj,
 
   assert(data);
   auto nl = static_cast<cnetlink *>(data);
-
-  // only enqueue nl msgs if not in stopped state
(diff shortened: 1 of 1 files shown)


src/netlink/cnetlink.cc
```diff
@@ -502,7 +502,7 @@ void cnetlink::nl_cb_v2(struct nl_cache *cache, struct nl_object *old_obj,
   auto nl = static_cast<cnetlink *>(data);
 
   // only enqueue nl msgs if not in stopped state
-  if (nl->state != NL_STATE_STOPPED)
+  if (nl->state == NL_STATE_RUNNING)
```
KanjiMonster, line 505: I don't think this works, the issue is we are receiving events while still in NL_STATE_STOPPED, so they never get added to `nl_objs`, so changing this to a check for `NL_STATE_RUNNING` will only delay it further.

Is there even any reason for this check? Wouldn't it make more sense to just unconditionally add them to `nl_objs`, and have the `NL_STATE_STOPPED` check in the code that processes the objects from `nl_objs`? That way we can't loose any event because we are in the wrong state.

PR #257, opened 2020-05-29
properly fix race in port tap device creation on of connection

KanjiMonster:
The fix in #254 was not enough due to a misunderstanding of how rofl-common works. So instead of trying to figure out the place where to enable netlink processing without missing messages, just enable it always.

## Description
Mostly see #254, but now enable when instantiating nbi_impl, instead of waiting for the signal that the switch is ready.

## Motivation and Context
See #254 

## How Has This Been Tested?
Two rounds of running the testing pipeline (with the pending carrier changes) on the ag7648, one round without the changes, build pipeline 9455 (currently running).

### DIFF
--- a/src/netlink/nbi_impl.cc
+++ b/src/netlink/nbi_impl.cc
@@ -17,6 +17,7 @@ nbi_impl::nbi_impl(std::shared_ptr<cnetlink> nl,
                    std::shared_ptr<tap_manager> tap_man)
     : nl(nl), tap_man(tap_man) {
   nl->set_tapmanager(tap_man);
+  nl->start();
 }
 
 nbi_impl::~nbi_impl() { nl->stop(); }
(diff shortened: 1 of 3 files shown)


src/netlink/nbi_impl.cc
```diff
@@ -31,12 +32,9 @@ void nbi_impl::register_switch(switch_interface *swi) noexcept {
 void nbi_impl::switch_state_notification(enum switch_state state) noexcept {
```
rubensfig, line 32: Do we need to maintain this switch_state_notifcation if we are starting the thread immediately? What would this function be doing after this patch?

KanjiMonster: On a first glance it does nothing anymore, but the state stuff is defined in `sai.h`, so I can't tell if this is wanted regardless (even if we don't react on it). Or if this has anything to do with SAI.

Issue #263, opened 2020-07-17
update IPv6 neighbors 


rubensfig:
## Expected Behavior
https://github.com/bisdn/basebox/commit/f40709880caa7c7ada0d891e9935e78d84933872#diff-e57a8238a1651a954280330119cafefcR471 

When updating the neighbour, baseboxd must check if the neighbors state is one of:
                    - permanent
                    - noarp                      
                    - reachable
                    - stale  
                    - none   
                    - incomplete
                    - delay  
                    - probe  
                    - failed 

In the case for incomplete neighbors, baseboxd must still install the rule to forward packets to the controller.
See https://github.com/bisdn/basebox/commit/0a5c7c1854ba0eb644b74fe36c136404f1e0cc6c#diff-3c368f43e25d51e19a158722ea158133R1547

## Actual Behavior

Currently there are no IPv6 neighbor updates being handled in baseboxd.

## Steps to Reproduce the Problem

bgp-ipv6 test

## Specifications
  - Version: baseboxd 1.6.0

PR #264, opened 2020-07-22
neigh: fix ipv6 neigh not being updated

rubensfig:
## Description
When updating the neighbour, baseboxd must check if the neighbors state is one of:
- permanent
- noarp
- reachable
- stale
- none
- incomplete
- delay
- probe
- failed

In the case for incomplete neighbors, baseboxd must still install the rule to forward packets to the controller.
See 0a5c7c1#diff-3c368f43e25d51e19a158722ea158133R1547

See #263 

## Motivation and Context

The L3 neighbor was not being updated when it moved out of the "failed" status. This PR fixes this behaviour.

## How Has This Been Tested?
-Tested with the bgp IPv6 test across two baseboxes

### DIFF
--- a/src/netlink/nl_l3.cc
+++ b/src/netlink/nl_l3.cc
@@ -676,19 +676,6 @@ int nl_l3::update_l3_neigh(struct rtnl_neigh *n_old, struct rtnl_neigh *n_new) {
   struct nl_addr *n_ll_old;
   struct nl_addr *n_ll_new;
 
-  int family = rtnl_neigh_get_family(n_new);
-  if (family == AF_INET6) {
-    struct nl_addr *addr = rtnl_neigh_get_dst(n_old);
-    rofl::caddress_in6 ipv6_dst = libnl_in6addr_2_rofl(addr, &rv);
-    if (rv < 0) {
-      LOG(ERROR) << __FUNCTION__ << ": could not parse addr " << addr;
-      return rv;
-    }
-
-    VLOG(1) << " new neigh " << ipv6_dst;
-    return 0;
-  }
-
   int ifindex = rtnl_neigh_get_ifindex(n_old);
   uint32_t port_id = nl->get_port_id(ifindex);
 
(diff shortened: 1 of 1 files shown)


src/netlink/nl_l3.cc
```diff
@@ -725,21 +711,37 @@ int nl_l3::update_l3_neigh(struct rtnl_neigh *n_old, struct rtnl_neigh *n_new) {
       return -EINVAL;
     }
 
-    struct nl_addr *addr = rtnl_neigh_get_dst(n_old);
-    rofl::caddress_in4 ipv4_dst = libnl_in4addr_2_rofl(addr, &rv);
+    int family = rtnl_neigh_get_family(n_new);
+    if (family == AF_INET6) {
+      struct nl_addr *addr = rtnl_neigh_get_dst(n_old);
+      rofl::caddress_in6 ipv6_dst = libnl_in6addr_2_rofl(addr, &rv);
```
DFritzsche, line 717: this could have a check of rv similar to ipv4 case

src/netlink/nl_l3.cc
```diff
@@ -725,21 +711,37 @@ int nl_l3::update_l3_neigh(struct rtnl_neigh *n_old, struct rtnl_neigh *n_new) {
...
 
-    // delete next hop
-    rv = sw->l3_unicast_host_remove(ipv4_dst);
-    if (rv < 0) {
-      LOG(ERROR) << __FUNCTION__
-                 << ": failed to remove l3 unicast host ipv4_dst" << ipv4_dst
-                 << "; rv=" << rv;
-      return rv;
+    } else if (family == AF_INET) {
+      struct nl_addr *addr = rtnl_neigh_get_dst(n_old);
```
DFritzsche, line 729: this is repetition, see ipv6 case

src/netlink/nl_l3.cc
```diff
@@ -725,21 +711,37 @@ int nl_l3::update_l3_neigh(struct rtnl_neigh *n_old, struct rtnl_neigh *n_new) {
       return -EINVAL;
     }
 
-    struct nl_addr *addr = rtnl_neigh_get_dst(n_old);
-    rofl::caddress_in4 ipv4_dst = libnl_in4addr_2_rofl(addr, &rv);
+    int family = rtnl_neigh_get_family(n_new);
+    if (family == AF_INET6) {
```
DFritzsche, line 715: can we have the ipv4 case before ipv6, as in every other function in this file?

PR #268, opened 2020-10-19
of-dpa: make sure we wait for the ofdpa-grpc connection

KanjiMonster:
## Description

Since we only establish the grpc connection to ofdpa-gprc after we
successfully created the openflow connection, there is a window where
the gprc connection isn't ready yet.

If it happens that we need to call any ofdpa grpc calls within this
window, these will fail, and the internal state becomes out of sync with
the linux state.

To avoid this, make grpc wait for the connection to be ready before
issuing rpc calls.

## Motivation and Context

Fixes bonding configuration braking when setup by systemd-networkd on
switches with fast CPUs.

## How Has This Been Tested?

Ran the the bonding tests that consistently failed twice successfully in a row, currently running a full testing pipeline on two ag5648s


rubensfig:
LGTM, merging when internal pipelines are done

PR #270, opened 2020-10-20
Add Support for l3 Addresses on trunk interfaces

rubensfig:
## Description

This PR aims to expand trunking support, allowing users to set l3 addresses on bond interfaces. Code added expands functionality in nl_bond class.

## Motivation and Context

Bonded interfaces also have support for IP addresses. Adding support to further L3 features, like routing daemons, requires that baseboxd correctly receives the bonding notifications, and translates this info to the switch.
 
## How Has This Been Tested?
Compiled with https://github.com/bisdn/rofl-ofdpa/pull/107.
Tests run:
  - bgp-ipv4-lacp
  - bgp-ipv4-active-backup
  - bgp-ipv6-lacp
  - bgp-ipv6-active-backup
  - ospf-ipv4-lacp
  - ospf-ipv4-active-backup
  - ospf-ipv6-lacp
  - ospf-ipv6-active-backup
  - eigrp-ipv4-lacp
  - eigrp-ipv4-active-backup
  - rip-ipv4-lacp
  - rip-ipv4-active-backup

### DIFF
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -321,6 +321,28 @@ std::set<uint32_t> cnetlink::get_bond_members_by_lag(rtnl_link *bond_link) {
   return bond->get_members(bond_link);
 }
 
+int cnetlink::get_l3_addrs(struct rtnl_link *link,
+                           std::deque<rtnl_addr *> *addresses) {
+  return l3->get_l3_addrs(link, addresses);
+}
+
+int cnetlink::add_l3_addr(struct rtnl_addr *a) {
+  switch (rtnl_addr_get_family(a)) {
+  case AF_INET:
+    return l3->add_l3_addr(a);
+  case AF_INET6:
+    return l3->add_l3_addr_v6(a);
+  default:
+    LOG(ERROR) << __FUNCTION__
+               << ": unsupported family=" << rtnl_addr_get_family(a)
+               << " for address=" << OBJ_CAST(a);
+    break;
+  }
+  return -EINVAL;
+}
+
+int cnetlink::del_l3_addr(struct rtnl_addr *a) { return l3->del_l3_addr(a); }
+
 struct rtnl_neigh *cnetlink::get_neighbour(int ifindex,
                                            struct nl_addr *a) const {
   assert(ifindex);
--- a/src/netlink/nl_bond.cc
+++ b/src/netlink/nl_bond.cc
@@ -131,7 +131,6 @@ int nl_bond::add_lag(rtnl_link *bond) {
       swi->lag_remove(lag_id);
   }
 
-  add_l3_address(bond);
 #endif
 
   return rv;
@@ -258,6 +257,7 @@ int nl_bond::add_lag_member(rtnl_link *bond, rtnl_link *link) {
   }
 
   // XXX FIXME check for vlan interfaces
+  add_l3_address(bond);
 #endif
 
   return rv;
--- a/src/netlink/nl_bond.cc
+++ b/src/netlink/nl_bond.cc
@@ -257,6 +257,11 @@ int nl_bond::add_lag_member(rtnl_link *bond, rtnl_link *link) {
   }
 
   // XXX FIXME check for vlan interfaces
+  // Adding an IP address here will ensure that every slave that is
+  // added will retry to add the address. It will not be written to the
+  // ASIC, but repeated messages will be seen
+  // This should be done in ::add_lag, but for currently unknown reasons
+  // this fails when the lag has no members yet. So keep it here for now.
   add_l3_address(bond);
 #endif
 
--- a/src/netlink/nl_bond.cc
+++ b/src/netlink/nl_bond.cc
@@ -68,6 +68,14 @@ int nl_bond::update_lag(rtnl_link *old_link, rtnl_link *new_link) {
   uint8_t o_mode, n_mode;
   uint32_t lag_id = nl->get_port_id(new_link);
 
+  if (lag_id == 0) {
+    rv = add_lag(new_link);
+    if (rv < 0)
+      return rv;
+
+    lag_id = rv;
+  }
+
   rv = rtnl_link_bond_get_mode(old_link, &o_mode);
   if (rv < 0) {
     VLOG(1) << __FUNCTION__ << ": failed to get mode for "
@@ -92,6 +100,8 @@ int nl_bond::update_lag(rtnl_link *old_link, rtnl_link *new_link) {
 
     return 0;
   }
+
+  add_l3_address(new_link);
 #endif
 
   return 0;
@@ -127,6 +137,8 @@ int nl_bond::add_lag(rtnl_link *bond) {
     if (lag_id != rv_emp.first->second)
       swi->lag_remove(lag_id);
   }
+
+  add_l3_address(bond);
 #endif
 
   return rv;
@@ -303,6 +315,11 @@ int nl_bond::remove_lag_member(rtnl_link *bond, rtnl_link *link) {
                  << OBJ_CAST(link);
   }
 
+  if (lm_rv->second.empty())
+    remove_l3_address(bond);
+
+  if (nl->is_bridge_interface(bond))
+    swi->ofdpa_stg_state_port_set(port_id, "forward");
 #endif
 
   return rv;
@@ -341,6 +358,48 @@ int nl_bond::update_lag_member(rtnl_link *old_slave, rtnl_link *new_slave) {
                                   new_state == 0);
 #endif
   return 0;
+}
+
+int nl_bond::add_l3_address(rtnl_link *link) {
+  int rv = 0;
+#ifdef HAVE_RTNL_LINK_BOND_GET_MODE
+  assert(link);
+
+  std::deque<rtnl_addr *> addresses;
+  nl->get_l3_addrs(link, &addresses);
+
+  for (auto i : addresses) {
+    LOG(INFO) << __FUNCTION__ << ": adding address=" << OBJ_CAST(i);
+
+    rv = nl->add_l3_addr(i);
+    if (rv < 0)
+      LOG(ERROR) << __FUNCTION__ << ":failed to add l3 address " << OBJ_CAST(i)
+                 << " to " << OBJ_CAST(link);
+  }
+  LOG(INFO) << __FUNCTION__ << ": added l3 addresses to bond "
+            << OBJ_CAST(link);
+
+#endif
+  return rv;
+}
+
+int nl_bond::remove_l3_address(rtnl_link *link) {
+  int rv = 0;
+#ifdef HAVE_RTNL_LINK_BOND_GET_MODE
+  assert(link);
+
+  std::deque<rtnl_addr *> addresses;
+  nl->get_l3_addrs(link, &addresses);
+
+  for (auto i : addresses) {
+    rv = nl->del_l3_addr(i);
+    if (rv < 0)
+      LOG(ERROR) << __FUNCTION__ << ":failed to remove l3 address from "
+                 << OBJ_CAST(link);
+  }
+
+#endif
+  return rv;
 }
 
 } // namespace basebox
--- a/src/netlink/nl_bond.cc
+++ b/src/netlink/nl_bond.cc
@@ -32,14 +32,7 @@ uint32_t nl_bond::get_lag_id(rtnl_link *bond) {
     return 0;
   }
 
-  auto rv_lm = lag_members.find(it->second);
-  if (rv_lm == lag_members.end()) {
-    VLOG(1) << ": no members in lag " << OBJ_CAST(bond);
-    return 0;
-  }
-
-  VLOG(3) << __FUNCTION__ << ": found id=" << rv_lm->first;
-  return rv_lm->first;
+  return it->second;
 }
 
 std::set<uint32_t> nl_bond::get_members(rtnl_link *bond) {
--- a/src/netlink/nl_l3.cc
+++ b/src/netlink/nl_l3.cc
@@ -251,18 +251,14 @@ int nl_l3::add_l3_addr(struct rtnl_addr *a) {
   }
 
   if (auto members = nl->get_bond_members_by_lag(link); !members.empty()) {
-    VLOG(1) << __FUNCTION__ << ": BOND configuring slaves";
+    VLOG(2) << __FUNCTION__ << ": configuring VLAN entry for bond slave "
+            << link;
 
     for (auto mem : members) {
       auto _link = nl->get_link_by_ifindex(nl->get_ifindex_by_port_id(mem));
       bool tagged = !!rtnl_link_is_vlan(link);
 
       rv = vlan->add_vlan(_link.get(), vid, tagged, vrf_id);
-
-      auto addr = rtnl_link_get_addr(_link.get());
-      rofl::caddress_ll mac = libnl_lladdr_2_rofl(addr);
-
-      rv = add_l3_termination(mem, vid, mac, AF_INET);
     }
   }
 
@@ -353,6 +349,18 @@ int nl_l3::add_l3_addr_v6(struct rtnl_addr *a) {
     if (rv < 0) {
       LOG(ERROR) << __FUNCTION__ << ": failed to add vlan id " << vid
                  << " (tagged=" << tagged << " to link " << OBJ_CAST(link);
+    }
+  }
+
+  if (auto members = nl->get_bond_members_by_lag(link); !members.empty()) {
+    VLOG(2) << __FUNCTION__ << ": configuring VLAN entry for bond slave "
+            << OBJ_CAST(link);
+
+    for (auto mem : members) {
+      auto _link = nl->get_link_by_ifindex(nl->get_ifindex_by_port_id(mem));
+      bool tagged = !!rtnl_link_is_vlan(link);
+
+      rv = vlan->add_vlan(_link.get(), vid, tagged);
     }
   }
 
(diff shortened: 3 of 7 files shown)


src/of-dpa/controller.cc
```diff
@@ -1085,12 +1085,21 @@ int controller::l3_egress_create(uint32_t port, uint16_t vid,
...
+                dpt.get_version(), _egress_interface_id, src_mac, dst_mac,
+                fm_driver.group_id_l2_trunk_interface(port, vid), false));
+    } else {
+      // TODO unfiltered interface
+      dpt.send_group_mod_message(
+          rofl::cauxid(0),
+          fm_driver.enable_group_l3_unicast(
+              dpt.get_version(), _egress_interface_id, src_mac, dst_mac,
+              fm_driver.group_id_l2_interface(port, vid), false));
+    }
```
KanjiMonster, line 1102: maybe something like this has a bit less code reuse:
```
   auto group_id;
  // TODO unfiltered interface
  if (nbi::get_port_type(port) == nbi::port_type_lag)
    group_id = fm_driver.group_id_l2_trunk_interface(port, vid);
  else
    group_id = fm_driver.group_id_l2_interface(port, vid);

  dpt.send_group_mod_message(
        rofl::cauxid(0),
        fm_driver.enable_group_l3_unicast(
        dpt.get_version(), _egress_interface_id, src_mac, dst_mac,
        group_id, false));
```

src/netlink/nl_l3.cc
```diff
@@ -356,6 +353,17 @@ int nl_l3::add_l3_addr_v6(struct rtnl_addr *a) {
...
   }
 
+  if (auto members = nl->get_bond_members_by_lag(link); !members.empty()) {
+    VLOG(1) << __FUNCTION__ << ": BOND configuring slaves";
+
+    for (auto mem : members) {
+      auto _link = nl->get_link_by_ifindex(nl->get_ifindex_by_port_id(mem));
+      bool tagged = !!rtnl_link_is_vlan(link);
+
+      rv = vlan->add_vlan(_link.get(), vid, tagged, vrf_id);
```
KanjiMonster, line 363: You never set `vrf_id` to anything but 0, so just pass 0 here.

src/netlink/nl_bond.cc
```diff
@@ -297,4 +297,21 @@ int nl_bond::update_lag_member(rtnl_link *old_slave, rtnl_link *new_slave) {
...
+  std::deque<rtnl_addr *> addresses;
+  auto bond = nl->get_link_by_ifindex(rtnl_link_get_master(link));
+  nl->get_l3_addr(bond.get(), &addresses);
+
+  for (auto i : addresses) {
+    nl->add_l3_addr(i);
+  }
+
+#endif
+  return rv;
```
KanjiMonster, line 314: This will break for `!HAVE_RTNL_LINK_BOND_GET_MODE`, since then `rv` is undefined. And even if `HAVE_RTNL_LINK_BOND_GET_MODE`  then `rv` is never set, so you return an undefined value. But this doesn't matter, since you never check the return value.

So either actually set `rv` and check it at the call locations, or just let this return `void`.

src/netlink/nl_bridge.cc
```diff
@@ -383,7 +383,7 @@ void nl_bridge::update_vlans(rtnl_link *old_link, rtnl_link *new_link) {
               for (auto mem : members) {
                 sw->egress_bridge_port_vlan_add(mem, vid, egress_untagged);
                 sw->ingress_port_vlan_add(mem, vid, new_br_vlan->pvid == vid);
-	      }
+              }
```
KanjiMonster, line 386: unrelated whitespace change

src/netlink/cnetlink.cc
```diff
@@ -1002,6 +1033,7 @@ void cnetlink::link_updated(rtnl_link *old_link, rtnl_link *new_link) noexcept {
   case LT_BOND_SLAVE:
     if (lt_new == LT_BOND_SLAVE) { // bond slave updated
       bond->update_lag_member(old_link, new_link);
+      bond->configure_l3_address(new_link);
```
KanjiMonster, line 1036: why do you need this? What happens if ip addresses get removed?

src/netlink/cnetlink.cc
```diff
@@ -1044,6 +1076,7 @@ void cnetlink::link_updated(rtnl_link *old_link, rtnl_link *new_link) noexcept {
                 << rtnl_link_get_name(new_link);
       rtnl_link *_bond = get_link(rtnl_link_get_master(new_link), AF_UNSPEC);
       bond->add_lag_member(_bond, new_link);
+      bond->configure_l3_address(new_link);
```
KanjiMonster, line 1079: Why do you need this?

src/netlink/nl_bond.cc
```diff
@@ -297,4 +297,21 @@ int nl_bond::update_lag_member(rtnl_link *old_slave, rtnl_link *new_slave) {
   return 0;
 }
 
+int nl_bond::configure_l3_address(rtnl_link *link) {
+#ifdef HAVE_RTNL_LINK_BOND_GET_MODE
+  assert(link);
+
+  int rv;
+  std::deque<rtnl_addr *> addresses;
+  auto bond = nl->get_link_by_ifindex(rtnl_link_get_master(link));
```
KanjiMonster, line 306: This could do with some error checking to make sure you don't pass on error values.

src/netlink/cnetlink.cc
```diff
@@ -1002,6 +1033,7 @@ void cnetlink::link_updated(rtnl_link *old_link, rtnl_link *new_link) noexcept {
   case LT_BOND_SLAVE:
     if (lt_new == LT_BOND_SLAVE) { // bond slave updated
       bond->update_lag_member(old_link, new_link);
+      bond->configure_l3_address(new_link);
     } else if (lt_new == LT_TUN) { // bond slave removed
       bond->remove_lag_member(old_link);
```
KanjiMonster, line 1038: If you add the ip addresses on enslave you should also remove them on deslave.

src/netlink/cnetlink.cc
KanjiMonster: Also maybe it would make more sense to move the configure_l3_address call being done within `nl_bond::{add,update,remove}_lag_member`, since this is part of the bond code?

src/of-dpa/controller.cc
```diff
@@ -1120,12 +1130,22 @@ int controller::l3_egress_update(uint32_t port, uint16_t vid,
...
+          rofl::cauxid(0),
+          fm_driver.enable_group_l3_unicast(dpt.get_version(), *l3_interface_id,
+                                            src_mac, dst_mac, group_id, true));
+    } else {
+      // TODO unfiltered interface
+      dpt.send_group_mod_message(
+          rofl::cauxid(0),
+          fm_driver.enable_group_l3_unicast(dpt.get_version(), *l3_interface_id,
+                                            src_mac, dst_mac, group_id, true));
+    }
```
KanjiMonster, line 1148: The call in both branches is now the same, you can drop the if else.

src/of-dpa/controller.cc
```diff
@@ -1085,12 +1085,22 @@ int controller::l3_egress_create(uint32_t port, uint16_t vid,
...
+                                 fm_driver.enable_group_l3_unicast(
+                                     dpt.get_version(), _egress_interface_id,
+                                     src_mac, dst_mac, group_id, false));
+    } else {
+      // TODO unfiltered interface
+      dpt.send_group_mod_message(rofl::cauxid(0),
+                                 fm_driver.enable_group_l3_unicast(
+                                     dpt.get_version(), _egress_interface_id,
+                                     src_mac, dst_mac, group_id, false));
+    }
```
KanjiMonster, line 1103: The call in both branches is now the same, you can drop the if else.

KanjiMonster (changes requested):
Please add explanation for your changes, and fix the mentioned issues.

src/netlink/cnetlink.cc
rubensfig: Not necessarily. Its a weird workflow, but a port can be enslaved - removed and another port enslaved again, and we would want to maintain the address on the bond.

KanjiMonster: but old_link and new_link will be both the same port; if they are different you would get a LT_BOND_SLAVE => LT_TUN for the one port and LT_TUN => LT_BOND_SLAVE for the other one.

KanjiMonster: But the address you took over from above were the addresses that are configured on the bond slave (AFAIU), if you remove the bond slave the addresses should not be configured on the bond anymore.

rubensfig: Ah, no. I was not configuring any addresses on the bond slave, since that is not a valid configuration.

What I was doing with this trigger, was checking on port enslavement, start a lookup for addresses on the bond interface, and if something was not configured then we write the address to the ASIC.

src/netlink/nl_bond.cc
```diff
@@ -257,6 +275,8 @@ int nl_bond::remove_lag_member(rtnl_link *bond, rtnl_link *link) {
 
   rv = swi->lag_remove_member(it->second, port_id);
   lag_members.erase(lm_rv);
+
+  remove_l3_address(bond);
```
KanjiMonster, line 279: what happens if I do

```
ip link add name bond1 type bond
ip link address add 10.0.01/24 dev bond1
ip link set port7 master bond1
ip link set port8 master bond1
ip link set port7 nomaster
```
Won't this remove the addresses in that case despite the bond still having members?

src/netlink/nl_l3.cc
```diff
@@ -258,11 +258,6 @@ int nl_l3::add_l3_addr(struct rtnl_addr *a) {
       bool tagged = !!rtnl_link_is_vlan(link);
 
       rv = vlan->add_vlan(_link.get(), vid, tagged, vrf_id);
-
-      auto addr = rtnl_link_get_addr(_link.get());
-      rofl::caddress_ll mac = libnl_lladdr_2_rofl(addr);
-
-      rv = add_l3_termination(mem, vid, mac, AF_INET);
```
KanjiMonster, line 265: Why do you remove this part?

src/netlink/nl_bond.cc
```diff
@@ -297,4 +317,42 @@ int nl_bond::update_lag_member(rtnl_link *old_slave, rtnl_link *new_slave) {
...
+  assert(link);
+
+  std::deque<rtnl_addr *> addresses;
+  nl->get_l3_addr(link, &addresses);
+
+  for (auto i : addresses) {
+    rv = nl->add_l3_addr(i);
+    if (rv < 0)
+      LOG(ERROR) << __FUNCTION__ << ":failed to add l3 address to "
+                 << OBJ_CAST(link);
```
KanjiMonster, line 332: Maybe also print the address itself?

src/netlink/nl_l3.cc
```diff
@@ -356,6 +351,17 @@ int nl_l3::add_l3_addr_v6(struct rtnl_addr *a) {
     }
   }
 
+  if (auto members = nl->get_bond_members_by_lag(link); !members.empty()) {
+    VLOG(1) << __FUNCTION__ << ": BOND configuring slaves";
```
KanjiMonster, line 355: This log line does not contain any useful information, you should also print the bond you are configuring slaves for

src/netlink/nl_l3.h
```diff
@@ -48,6 +48,8 @@ class nl_l3 {
   int update_l3_neigh(struct rtnl_neigh *n_old, struct rtnl_neigh *n_new);
   int del_l3_neigh(struct rtnl_neigh *n);
 
+  int get_l3_addr(struct rtnl_link *link, std::deque<rtnl_addr *> addresses);
```
KanjiMonster, line 51: addr -> addrs? (since you are returning multiple

src/netlink/nl_bond.cc
```diff
@@ -220,6 +237,7 @@ int nl_bond::add_lag_member(rtnl_link *bond, rtnl_link *link) {
   }
 
   // XXX FIXME check for vlan interfaces
+  add_l3_address(bond);
```
KanjiMonster, line 240: Shouldn't you only do this if you called `::add_lag`?

src/netlink/nl_l3.cc
```diff
@@ -356,6 +351,17 @@ int nl_l3::add_l3_addr_v6(struct rtnl_addr *a) {
...
+
+    for (auto mem : members) {
+      auto _link = nl->get_link_by_ifindex(nl->get_ifindex_by_port_id(mem));
+      bool tagged = !!rtnl_link_is_vlan(link);
+
+      rv = vlan->add_vlan(_link.get(), vid, tagged, 0);
```
KanjiMonster, line 361: Actually 0 is the default passed when omitting the argument, so you can just drop it
```suggestion
      rv = vlan->add_vlan(_link.get(), vid, tagged);
```

KanjiMonster (changes requested):

src/netlink/nl_l3.cc
rubensfig: We dont need to add the bond_slave termination MACs anymore, there is the bond interface TMAC entry.

src/netlink/nl_bond.cc
```diff
@@ -323,12 +325,12 @@ int nl_bond::add_l3_address(rtnl_link *link) {
...
 
   std::deque<rtnl_addr *> addresses;
-  nl->get_l3_addr(link, &addresses);
+  nl->get_l3_addrs(link, &addresses);
 
   for (auto i : addresses) {
     rv = nl->add_l3_addr(i);
     if (rv < 0)
-      LOG(ERROR) << __FUNCTION__ << ":failed to add l3 address to "
+      LOG(ERROR) << __FUNCTION__ << ":failed to add l3 address " << i << " to "
```
KanjiMonster, line 333: Does this work without OBJ_CAST'ing the address (i)?

src/netlink/cnetlink.cc
```diff
@@ -321,6 +321,26 @@ std::set<uint32_t> cnetlink::get_bond_members_by_lag(rtnl_link *bond_link) {
...
+}
+
+int cnetlink::add_l3_addr(struct rtnl_addr *a) {
+  switch (rtnl_addr_get_family(a)) {
+  case AF_INET:
+    return l3->add_l3_addr(a);
+  case AF_INET6:
+    return l3->add_l3_addr_v6(a);
+  default:
+    LOG(ERROR) << __FUNCTION__ << ": unsupported family ";
```
KanjiMonster, line 336: please print the address and family here so we know what isn't supported (and maybe deduce where it came from)

src/netlink/nl_bond.cc
```diff
@@ -67,6 +67,22 @@ int nl_bond::update_lag(rtnl_link *old_link, rtnl_link *new_link) {
   uint8_t o_mode, n_mode;
   uint32_t lag_id = nl->get_port_id(new_link);
 
+  if (lag_id == 0) {
+    int master_id = rtnl_link_get_master(new_link);
+    auto master = nl->get_link(master_id, AF_UNSPEC);
+
+    if (master == nullptr) {
+      LOG(ERROR) << __FUNCTION__ << ": failed to get master " << master;
```
KanjiMonster, line 75: did you intend to print `master_id` or `new_link` here? `master` can only be a nullptr at this point.

src/netlink/nl_bond.cc
```diff
@@ -297,4 +319,42 @@ int nl_bond::update_lag_member(rtnl_link *old_slave, rtnl_link *new_slave) {
...
+#ifdef HAVE_RTNL_LINK_BOND_GET_MODE
+  assert(link);
+
+  std::deque<rtnl_addr *> addresses;
+  nl->get_l3_addrs(link, &addresses);
+
+  for (auto i : addresses) {
+    rv = nl->del_l3_addr(i);
+    if (rv < 0)
+      LOG(ERROR) << __FUNCTION__ << ":failed to add l3 address to "
```
KanjiMonster, line 352:
```suggestion
      LOG(ERROR) << __FUNCTION__ << ":failed to remove l3 address from "
```

And maybe print the address?

src/netlink/nl_l3.cc
```diff
@@ -356,6 +351,18 @@ int nl_l3::add_l3_addr_v6(struct rtnl_addr *a) {
...
+    VLOG(2) << __FUNCTION__ << ": configuring VLAN entry for bond slave "
+            << link;
+
+    for (auto mem : members) {
+      auto _link = nl->get_link_by_ifindex(nl->get_ifindex_by_port_id(mem));
+      bool tagged = !!rtnl_link_is_vlan(link);
+
+      rv = vlan->add_vlan(_link.get(), vid, tagged);
+    }
+  }
```
KanjiMonster, line 364: 1. This whole block should be in the `if (!is_loopback && !nl->is_bridge_interface())` branch.
2. Don't you need to add a similar block to `nl_l3::del_l3_addr` to remove it?
3. bond interfaces should probably be handled in add/remove vlan (=> https://github.com/bisdn/basebox/commit/7136fe42f8e6fe6795e4e22f77572092d4c87f7a )

src/of-dpa/controller.cc
```diff
@@ -1120,12 +1122,14 @@ int controller::l3_egress_update(uint32_t port, uint16_t vid,
 
   try {
     rofl::crofdpt &dpt = set_dpt(dptid, true);
-    // TODO unfiltered interface
```
KanjiMonster, line 1123: This TODO should stay.

KanjiMonster (changes requested):

KanjiMonster (approved):

src/netlink/nl_l3.cc
```diff
@@ -931,16 +960,6 @@ int nl_l3::add_l3_egress(int ifindex, const uint16_t vid,
...
-  } else {
-    for (auto i = it.first; i != it.second; ++i) {
-      if (i->first == l3_if_tuple) {
-        if (l3_interface_id)
-          *l3_interface_id = i->second.l3_interface_id;
-        i->second.refcnt++;
-        rv = 0;
-        break;
-      }
-    }
```
KanjiMonster, line 943: Please add an explanation for this change to the commit message.

src/netlink/nl_bond.cc
```diff
@@ -289,20 +302,14 @@ int nl_bond::remove_lag_member(rtnl_link *bond, rtnl_link *link) {
   }
 
   rv = swi->lag_remove_member(it->second, port_id);
-  lag_members.erase(lm_rv);
 
-  if (nl->is_bridge_interface(bond)) {
-    swi->ofdpa_stg_state_port_set(port_id, "forward");
-
-    auto br_link = nl->get_link(rtnl_link_get_ifindex(bond), AF_BRIDGE);
-    rv = nl->unset_bridge_port_vlan_tpid(br_link);
```
KanjiMonster, line 298: Why did you drop this?

src/netlink/nl_bond.cc
KanjiMonster: And in that case, shouldn't have ::add_lag already taken care of this?

KanjiMonster (changes requested):

PR #274, opened 2020-11-04
implement stp for bond interfaces

KanjiMonster:
Implement STP state support for bonded interfaces by passing on the state to the bonded interfaces.

## Description

The switch ASIC only supports setting STP state for physical ports, so we need to make sure that we propagate any STP state changes to the bonded interfaces. This also fixes an issue in get_port_id() that wasn't working with bridge interfaces for bond interfaces, only ones for tap interfaces.

## Motivation and Context

STP is essential in many use cases.

## How Has This Been Tested?

Ran a pipeline with the changes applied on the ag5648s.

### DIFF
--- a/src/netlink/nl_bond.cc
+++ b/src/netlink/nl_bond.cc
@@ -6,6 +6,7 @@
 
 #include <cassert>
 #include <glog/logging.h>
+#include <linux/if_bridge.h>
 #include <netlink/route/link.h>
 #include <netlink/route/link/bonding.h>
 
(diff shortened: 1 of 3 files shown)


src/netlink/nl_bond.cc
```diff
@@ -216,6 +217,32 @@ int nl_bond::add_lag_member(rtnl_link *bond, rtnl_link *link) {
...
+        break;
+      case BR_STATE_LISTENING:
+        state = "listen";
+        break;
+      case BR_STATE_LEARNING:
+        state = "learn";
+        break;
+      default:
+        VLOG(1) << __FUNCTION__ << ": stp state change not supported";
+        return rv;
```
DFritzsche, line 242: Are you sure to return rv here?

DFritzsche (approved):

PR #283, opened 2020-11-23
controller: expose error message

rubensfig:
Signed-off-by: Rubens Figueiredo <rubens.figueiredo@bisdn.de>

## Description
This PR exposes error messages from the ASIC through logging.

## Motivation and Context
Exposing the error messages we received is necessary to debug and see what the system is correctly working.

### DIFF
--- a/src/of-dpa/controller.cc
+++ b/src/of-dpa/controller.cc
@@ -252,11 +252,10 @@ void controller::handle_error_message(rofl::crofdpt &dpt,
                                       const rofl::cauxid &auxid,
                                       rofl::openflow::cofmsg_error &msg) {
   VLOG(1) << __FUNCTION__ << ": dpt=" << dpt << " on auxid=" << auxid;
-  VLOG(3) << __FUNCTION__ << ": dpid=" << dpt.get_dpid()
+  VLOG(1) << __FUNCTION__ << ": dpid=" << dpt.get_dpid()
           << " pkt received: " << std::endl
           << msg;
-
-  LOG(WARNING) << __FUNCTION__ << ": not implemented";
+  LOG(WARNING) << __FUNCTION__ << ": " << (((uint32_t)msg.get_err_type() << 16) | msg.get_err_code());
 }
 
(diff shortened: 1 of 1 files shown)


src/of-dpa/controller.cc
```diff
@@ -256,7 +256,7 @@ void controller::handle_error_message(rofl::crofdpt &dpt,
           << " pkt received: " << std::endl
           << msg;
 
-  LOG(WARNING) << __FUNCTION__ << ": not implemented";
+  LOG(WARNING) << __FUNCTION__ << msg;
```
KanjiMonster, line 259: Can we avoid printing the whole packet and just the message itself? The whole packet would be rather something for VLOG.

src/of-dpa/controller.cc
```diff
@@ -252,11 +252,11 @@ void controller::handle_error_message(rofl::crofdpt &dpt,
                                       const rofl::cauxid &auxid,
                                       rofl::openflow::cofmsg_error &msg) {
   VLOG(1) << __FUNCTION__ << ": dpt=" << dpt << " on auxid=" << auxid;
-  VLOG(3) << __FUNCTION__ << ": dpid=" << dpt.get_dpid()
+  VLOG(1) << __FUNCTION__ << ": dpid=" << dpt.get_dpid()
           << " pkt received: " << std::endl
           << msg;
 
-  LOG(WARNING) << __FUNCTION__ << ": not implemented";
+  LOG(ERROR) << __FUNCTION__ << msg.get_err_code();
```
KanjiMonster, line 259: Why did you change the severity? Looks like there is no way of getting the string representation of the error without the full contents ... .

You also probably want to get the err_type as well, since this is also part of the error id:

```
 LOG(WARNING) << __FUNCTION__ << ": " << ((uint32_t)msg.get_err_type() << 16) | msg.get_err_code());
```

KanjiMonster (changes requested):

KanjiMonster (approved):

PR #287, opened 2020-12-10
Extend support for VRFs on bonded interfaces

rubensfig:
## Description
This PR aims to extend VRF support.  We have centralized all VRF handling on the nl_vlan classes, which then provides a clean way to configure VRFs on all interfaces, removing the specific order limitation that was present until the current PR. 

We have started tracking all of the VLAN - VRF mappings, so that it is easier to lookup the ids when we need to configure this.

## Motivation and Context
This Pr extends the work done to support bond interfaces.

## How Has This Been Tested?
Ran pipelines including the trunk-vrf tests.

### DIFF
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -1103,9 +1103,10 @@ void cnetlink::link_updated(rtnl_link *old_link, rtnl_link *new_link) noexcept {
               << OBJ_CAST(
                      get_link_by_ifindex(rtnl_link_get_master(new_link)).get());
     // Lookup l3 addresses to delete
+    // We can delete the addresses on the interface because we will later receive a
+    // notification readding the addresses, that time with the correct VRF
     std::deque<rtnl_addr *> addresses;
     get_l3_addrs(old_link, &addresses);
-    // delete l3 addresses
     for (auto i : addresses)
       l3->del_l3_addr(i);
 
@@ -1117,9 +1118,11 @@ void cnetlink::link_updated(rtnl_link *old_link, rtnl_link *new_link) noexcept {
                 << OBJ_CAST(old_link);
 
       // Lookup l3 addresses to delete
+      // We can delete the addresses on the interface because we will later receive a
+      // notification readding the addresses, that time with the correct VRF
       std::deque<rtnl_addr *> addresses;
       get_l3_addrs(old_link, &addresses);
-     // delete l3 addresses no longer associated with vrf 
+      // delete l3 addresses no longer associated with vrf 
       for (auto i: addresses)
           l3->del_l3_addr(i);
       vlan->vrf_detach(old_link, new_link);
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -1105,13 +1105,16 @@ void cnetlink::link_updated(rtnl_link *old_link, rtnl_link *new_link) noexcept {
     vlan->vrf_attach(old_link, new_link);
     break;
   case LT_VRF_SLAVE:
-    if (lt_new != LT_VLAN) {
-      VLOG(1) << __FUNCTION__ << ": ignoring update of VRF SLAVE interface";
-      return;
+    if (lt_new == LT_VLAN) {
+      LOG(INFO) << __FUNCTION__ << ": freed vrf slave interface "
+                << OBJ_CAST(old_link);
+      vlan->vrf_detach(old_link, new_link);
+    } else if (lt_new == LT_VRF_SLAVE) {
+      LOG(INFO) << __FUNCTION__ << ": link updated " << OBJ_CAST(new_link);
+      vlan->vrf_attach(old_link, new_link);
     }
-    LOG(INFO) << __FUNCTION__ << ": freed vrf slave interface "
-              << OBJ_CAST(old_link);
-    l3->vrf_detach(old_link, new_link);
+    // TODO handle LT_TAP/LT_BOND, currently unimplemented
+    // happens when a bond/tap interface enslaved directly to a VRF and you set nomaster
     break;
   case LT_TUN:
     if (lt_new == LT_BOND_SLAVE) {
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -1093,7 +1093,7 @@ void cnetlink::link_updated(rtnl_link *old_link, rtnl_link *new_link) noexcept {
       }
     }
     break;
-  case LT_VLAN:
+  case LT_VLAN: {
     if (lt_new != LT_VRF_SLAVE) {
       VLOG(1) << __FUNCTION__ << ": ignoring update of VLAN interface";
       break;
@@ -1102,9 +1102,16 @@ void cnetlink::link_updated(rtnl_link *old_link, rtnl_link *new_link) noexcept {
               << rtnl_link_get_name(new_link) << " to vrf "
               << OBJ_CAST(
                      get_link_by_ifindex(rtnl_link_get_master(new_link)).get());
+    // Lookup l3 addresses to delete
+    std::deque<rtnl_addr *> addresses;
+    get_l3_addrs(old_link, &addresses);
+    // delete l3 addresses
+    for (auto i : addresses)
+      l3->del_l3_addr(i);
+
     vlan->vrf_attach(old_link, new_link);
-    break;
-  case LT_VRF_SLAVE:
+    } break;
+  case LT_VRF_SLAVE: {
     if (lt_new == LT_VLAN) {
       LOG(INFO) << __FUNCTION__ << ": freed vrf slave interface "
                 << OBJ_CAST(old_link);
@@ -1122,7 +1129,7 @@ void cnetlink::link_updated(rtnl_link *old_link, rtnl_link *new_link) noexcept {
     }
     // TODO handle LT_TAP/LT_BOND, currently unimplemented
     // happens when a bond/tap interface enslaved directly to a VRF and you set nomaster
-    break;
+    } break;
   case LT_TUN:
     if (lt_new == LT_BOND_SLAVE) {
       // XXX link enslaved
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -1108,6 +1108,13 @@ void cnetlink::link_updated(rtnl_link *old_link, rtnl_link *new_link) noexcept {
     if (lt_new == LT_VLAN) {
       LOG(INFO) << __FUNCTION__ << ": freed vrf slave interface "
                 << OBJ_CAST(old_link);
+
+      // Lookup l3 addresses to delete
+      std::deque<rtnl_addr *> addresses;
+      get_l3_addrs(old_link, &addresses);
+     // delete l3 addresses no longer associated with vrf 
+      for (auto i: addresses)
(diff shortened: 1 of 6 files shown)


src/netlink/cnetlink.cc
```diff
@@ -1075,16 +1102,19 @@ void cnetlink::link_updated(rtnl_link *old_link, rtnl_link *new_link) noexcept {
...
-    LOG(INFO) << __FUNCTION__ << ": freed vrf slave interface "
-              << OBJ_CAST(old_link);
-    l3->vrf_detach(old_link, new_link);
+    if (lt_new == LT_VLAN) {
+      LOG(INFO) << __FUNCTION__ << ": freed vrf slave interface "
+                << OBJ_CAST(old_link);
+      vlan->vrf_detach(old_link, new_link);
+    } else if (lt_new == LT_VRF_SLAVE) {
+      LOG(INFO) << __FUNCTION__ << ": link updated " << OBJ_CAST(new_link);
+      vlan->vrf_attach(old_link, new_link);
```
KanjiMonster, line 1114: this looks like a fix that should be a separate commit

src/netlink/cnetlink.cc
```diff
@@ -1075,16 +1102,19 @@ void cnetlink::link_updated(rtnl_link *old_link, rtnl_link *new_link) noexcept {
...
+    } // check for unexpected cases
```
KanjiMonster, line 1115: is here a TODO missing, or did you intend to add something here?

src/netlink/nl_l3.cc
```diff
@@ -383,7 +378,7 @@ int nl_l3::del_l3_addr(struct rtnl_addr *a) {
   int rv = 0;
   int family = rtnl_addr_get_family(a);
   struct rtnl_link *link = rtnl_addr_get_link(a);
-  uint16_t vrf_id = get_vrf_table_id(link);
+  uint16_t vrf_id = nl->get_vrf_table_id(link);
```
KanjiMonster: Any reason for not doing
```
  uint32_t vrf_id = vlan->get_vrf_id(vid, link);
```
?

src/netlink/nl_l3.cc
```diff
@@ -523,7 +518,7 @@ int nl_l3::add_l3_neigh_egress(struct rtnl_neigh *n, uint32_t *l3_interface_id,
     // the egress entries for the Bridge SVIs are the ports
     // attached to the bridge
     auto lladdr = rtnl_neigh_get_lladdr(n);
-    *vrf_id = get_vrf_table_id(link.get());
+    *vrf_id = nl->get_vrf_table_id(link.get());
```
KanjiMonster: similar here
```
  *vrf_id = vlan->get_vrf_id(vid, link);
```
?

src/netlink/nl_vlan.cc
```diff
@@ -346,4 +355,79 @@ uint16_t nl_vlan::get_vid(rtnl_link *link) {
   return vid;
 }
 
+uint16_t nl_vlan::get_vrf_id(uint16_t vid, rtnl_link *link) {
+  auto vlanmap = vlan_vrf.find(vid);
+  if (vlanmap == vlan_vrf.end()) {
+    auto id = nl->get_vrf_table_id(link);
```
KanjiMonster, line 361: Do you really need this? Shouldn't the map have the vrf-id since you add it in attach?

src/netlink/nl_vlan.cc
```diff
@@ -346,4 +355,79 @@ uint16_t nl_vlan::get_vid(rtnl_link *link) {
...
+  }
+
+  LOG(INFO) << __FUNCTION__ << ": map vlan=" << vid << " vrf=" << vrf;
+
+  // delete old entries
+  auto fdb_entries = nl->search_fdb(vid, nullptr);
+  for (auto entry : fdb_entries) {
+    auto link = nl->get_link_by_ifindex(rtnl_neigh_get_ifindex(entry));
+
+    remove_ingress_vlan(nl->get_port_id(link.get()), vid, true, 0);
```
KanjiMonster, line 411: don't you need to re-add the vlan entries again with the vrf set?

src/netlink/nl_vlan.cc
```diff
@@ -346,4 +355,79 @@ uint16_t nl_vlan::get_vid(rtnl_link *link) {
...
+    LOG(ERROR) << __FUNCTION__ << ": no VRF found for vlan=" << vid;
+
+  }
+
+  // delete old entries
+  auto fdb_entries = nl->search_fdb(vid, nullptr);
+  for (auto entry : fdb_entries) {
+    auto link = nl->get_link_by_ifindex(rtnl_neigh_get_ifindex(entry));
+
+    remove_ingress_vlan(nl->get_port_id(link.get()), vid, true, vlanmap->second);
```
KanjiMonster, line 428: similar, don't you need to readd them again with vrf unset/0?

KanjiMonster (changes requested):
The first commit has "bonding:" int the title, but has no changes to ::nl_bond. Did you forget to add the changes?

The second commit isn't related to the title of the PR, so should go in a separate PR.

src/netlink/nl_vlan.cc
rubensfig: We can just wait that further updates to the interfaces set these entries, so we only need to remove the old state.

KanjiMonster: Can you please show these updates happening?

e.g. will if e.g. swbridge.10 has a vrf set, will

```
ip link set swbridge.10 nomaster
```
both trigger the removal and whatever is needed to re-add the vlans?

src/netlink/nl_vlan.cc
rubensfig: We do `uint16_t vrf = get_vrf_id(vid, new_link);` on vrf attach. If the id is not on the map *yet* we need to parse get the table id and add it to the map.

Maybe the function is misnamed, since it does add the vrf id when it does not find it in the map.

KanjiMonster: Maybe let vrf_attach call get_vrf_table_id instead of get_vrf_id, and drop the get_vrf_table_id usage from here. Only vrf_attach should extract the vrf_id from the netlink data, everything else should use the map. vrf_attach may even complain if the map already contains a different value for the vlan.

Could it happen that code starts using a vrf (because get_vrf_table_id() returns it) before we handled the attach in vrf_attach? That might lead to weird subtle bugs.

src/netlink/cnetlink.cc
```diff
@@ -1108,6 +1108,13 @@ void cnetlink::link_updated(rtnl_link *old_link, rtnl_link *new_link) noexcept {
     if (lt_new == LT_VLAN) {
       LOG(INFO) << __FUNCTION__ << ": freed vrf slave interface "
                 << OBJ_CAST(old_link);
+
+      // Lookup l3 addresses to delete
+      std::deque<rtnl_addr *> addresses;
+      get_l3_addrs(old_link, &addresses);
+     // delete l3 addresses no longer associated with vrf 
+      for (auto i: addresses)
+          l3->del_l3_addr(i);
```
KanjiMonster, line 1117: These IP addresses are still assigned to the vlan interface, are they not? Should their entries be just updated to not use reference the vrf anymore instead of removed? And vice versa on the attach path? Might make sense to move this to `vrf_{at,de}tach` then.

KanjiMonster (commented):
LGTM

src/netlink/cnetlink.cc
rubensfig: The ip addresses will be readded again shortly after deleting the IP addresses on the link.

```
Dec 16 11:16:41 accton-as4610 baseboxd[1289]: W1216 11:16:41.362574  1298 nl_bridge.cc:776] learn_source_mac: got packet on unconfigured port
Dec 16 11:16:41 accton-as4610 baseboxd[1289]: I1216 11:16:41.962591  1298 cnetlink.cc:1109] link_updated: freed vrf slave interface swbridge.100 ether 4a:0a:73:9d:bc:87 master red <broadcast,multicast,up,running,lowerup> slave-of swbri>
Dec 16 11:16:41 accton-as4610 baseboxd[1289]: I1216 11:16:41.962796  1298 nl_l3.cc:474] get_l3_addrs: searching l3 addresses from interface=swbridge.100 ether 4a:0a:73:9d:bc:87 master red <broadcast,multicast,up,running,lowerup> slave->
...
Dec 16 11:16:41 accton-as4610 baseboxd[1289]: I1216 11:16:41.971890  1298 nl_l3.cc:1260] get_neighbours_of_route: ifindex=62 gw=10.0.100.0/24
Dec 16 11:16:41 accton-as4610 baseboxd[1289]: I1216 11:16:41.971930  1298 nl_l3.cc:1268] get_neighbours_of_route: get neigh of nh_addr=10.0.100.0/24
Dec 16 11:16:41 accton-as4610 baseboxd[1289]: I1216 11:16:41.971967  1298 nl_l3.cc:1535] add_l3_unicast_route: got unresolved nh ifindex=62, nh=10.0.100.0/24
Dec 16 11:16:41 accton-as4610 baseboxd[1289]: I1216 11:16:41.972020  1298 nl_l3.cc:1545] add_l3_unicast_route: got 0 neighbours (1 unresolved next hops)
```

rubensfig: We do need to process the same in VRF attach, so that this state is avoided. Obtained by running https://github.com/bisdn/basebox/blob/master/examples/routing/VRFs/01-controller and adding first the IP addresses and then configuring the VRF.

```
Table ID 30 (Unicast Routing):   Retrieving all entries. Max entries = 32768, Current entries = 4.
--  etherType = 0x0800 vrf:mask = 0x0000:0x0000 dstIp4 = 10.0.100.0/255.255.255.0 dstIp6 = ::/:: | GoTo = 60 (ACL Policy) outPort = CONTROLLER (Reserved)  | priority = 2 hard_time = 0 idle_time = 0 cookie = 16
--  etherType = 0x0800 vrf:mask = 0x0000:0x0000 dstIp4 = 10.0.200.0/255.255.255.0 dstIp6 = ::/:: | GoTo = 60 (ACL Policy) outPort = CONTROLLER (Reserved)  | priority = 2 hard_time = 0 idle_time = 0 cookie = 19
--  etherType = 0x0800 vrf:mask = 0x3039:0xffff dstIp4 = 10.0.100.0/255.255.255.0 dstIp6 = ::/:: | GoTo = 60 (ACL Policy) outPort = CONTROLLER (Reserved)  | priority = 2 hard_time = 0 idle_time = 0 cookie = 26
--  etherType = 0x0800 vrf:mask = 0x3039:0xffff dstIp4 = 10.0.200.0/255.255.255.0 dstIp6 = ::/:: | GoTo = 60 (ACL Policy) outPort = CONTROLLER (Reserved)  | priority = 2 hard_time = 0 idle_time = 0 cookie = 33
```

rubensfig: I would also process the IP address not on VRF_{at,de}tach, but on cnetlink, both due to a missing interface to the l3 class, and also because of code workflow.

KanjiMonster: Can you please add a comment in the code there why it is okay to delete the addresses and we cannot just update the flows?

KanjiMonster (approved):
LGTM now

PR #289, opened 2020-12-16
Whitespace fix

rubensfig:


rubensfig:
Depends on #287  and #288

PR #293, opened 2021-01-22
baseboxd: implement per vlan spanning tree functionality

rubensfig:
## Motivation and Context

This PR aims to add Per VLAN spanning tree state functionality to baseboxd, required by protocols like MSTP and PVST. 

This is achieved by tracking the VLAN Database changes in the kernel, that stores VLAN and STP states in all Linux interfaces(tap/bridge/bond/...). The kernel reports changes to the database via NETLINK messages RTM_GET/NEW/DELVLAN, and baseboxd is able to cache these messages via libnl. See: https://github.com/rubensfig/libnl/tree/bridge_vlandb

Setting the state on the ASIC is done via the gRPC connection, with the model from https://github.com/bisdn/basebox-protobuf/pull/9.

## Description

This PR adds receiving and parsing of VLAN Database STP states to nl_bridge, where all bridging related events are handled. We obtain the Kernel changes to the VLAN DB via NETLINK, and then process the resulting bridge vlan object to reflect the state on the ASIC. 

We streamlined behaviour in global STP state handling in the ASIC by setting the first STG (spanning tree group), enabled with all VLANS and all ports. When the RTM_NEWVLAN message is received we obtain the interface index of a port, the port state and VLAN, and map that VLAN id onto a new or exisiting STG id, therefore allowing us to individually set the port's state on a per VLAN/per port basis. 

## How Has This Been Tested?
Ran internal testing pipelines. Tested with systemd and iproute2

### DIFF
--- a/src/netlink/nl_bridge.h
+++ b/src/netlink/nl_bridge.h
@@ -89,7 +89,7 @@ struct bridge_stp_states {
   uint8_t get_global_state(int port_id) {
     auto it = gl_states.find(port_id);
 
-    return (it == gl_states.end()) ? 0 : it->second;
+    return (it == gl_states.end()) ? -EINVAL : it->second;
   }
 
   uint8_t get_pvlan_state(int port_id, uint16_t vid) {
@@ -97,12 +97,12 @@ struct bridge_stp_states {
 
     // no vlan found
     if (it == pv_states.end())
-      return 0;
+      return -EINVAL;
 
     // no port found
     auto pv_state = it->second.find(port_id);
     if (pv_state == it->second.end())
-      return 0;
+      return -EINVAL;
 
     return pv_state->second;
   }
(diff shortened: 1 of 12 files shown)


src/netlink/nl_bridge.h
```diff
@@ -127,8 +127,8 @@ struct bridge_stp_states {
     int g_state = get_global_state(port_id);
     int pv_state = get_pvlan_state(port_id, vid);
 
-    if (pv_state == 0)
-      return 0;
+    if (pv_state < 0)
+      return pv_state;
```
KanjiMonster, line 131: you probably want to return the g_state here

src/netlink/nl_bridge.h
```diff
@@ -127,8 +127,8 @@ struct bridge_stp_states {
...
 
     if (g_state == BR_STATE_BLOCKING || pv_state == BR_STATE_BLOCKING)
       return g_state;
```
KanjiMonster, line 134: you need to return BR_STATE_BLOCKING here

PR #298, opened 2021-02-26
nl_l3: update Link Layer Egress Entries

rubensfig:
## Description
This PR deals with updating L3 egress entries. 
There are L3 egress entries being removed on IPv6 Link Layer addresses neigh removal. This is a mismatch since we are not refcounting ipv6 ll addresses.

Additionally, we have to track nexthop changes, more specifically updating routes over a nexthop when it turns unreachable. This is done via registering a route and its nexthop on addition, and waiting on this object until the corresponding nexthop gets deleted via del_l3_neigh. We then rewrite the route to controller, which is a mismatch to Linux state. Linux "wipes" the route from the kernel state, but we do not receive a specific notification for this event.

## How Has This Been Tested?
Internal static-route test on the 4610.

### DIFF
--- a/src/netlink/nl_l3.cc
+++ b/src/netlink/nl_l3.cc
@@ -588,7 +588,7 @@ int nl_l3::del_l3_neigh_egress(struct rtnl_neigh *n) {
 
   // remove egress group reference
   int rv = del_l3_egress(ifindex, vid, s_mac, d_mac);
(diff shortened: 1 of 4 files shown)


KanjiMonster:
Please provide a proper description of the issue and why your commit fixes it

hilmarm:
lgtm

PR #301, opened 2021-03-11
nl_l3: add neigh search and update

rubensfig:
## Description
In the case that a certain neighbor is pointing to another egress entry, on address addition we must rewrite the nexthop to point to controller

## Motivation and Context
This use case is seen for example in VRRP.

## How Has This Been Tested?
Ran keepalived test

### DIFF
--- a/src/netlink/nl_l3.cc
+++ b/src/netlink/nl_l3.cc
@@ -358,7 +358,7 @@ int nl_l3::add_l3_addr_v6(struct rtnl_addr *a) {
   bool update = false;
   if (std::list<struct rtnl_neigh *> neigh;
       search_neigh_cache(rtnl_addr_get_ifindex(a), addr, AF_INET6, &neigh) == 0) {
-    VLOG(2) << __FUNCTION__ << ": list neigh size " << neigh.size();
+    VLOG(3) << __FUNCTION__ << ": list neigh size " << neigh.size();
     if (neigh.size() > 0)
       update = true;
   }
@@ -886,8 +886,6 @@ int nl_l3::del_l3_neigh(struct rtnl_neigh *n) {
   for (auto i : link_addresses) {
     if (struct nl_addr *link_addr = rtnl_addr_get_local(i);
         nl_addr_cmp(addr, link_addr) == 0) {
-      VLOG(1) << __FUNCTION__
-              << ": address configured on the interface, skipping";
       skip_addr_remove = true;
       break;
     }
--- a/src/netlink/nl_l3.cc
+++ b/src/netlink/nl_l3.cc
@@ -129,6 +129,26 @@ int nl_l3::init() noexcept {
   return 0;
 }
 
+int nl_l3::search_neigh_cache(int ifindex, struct nl_addr *addr, std::list<struct rtnl_neigh *> *neigh) {
+  std::unique_ptr<struct rtnl_neigh, void (*)(rtnl_neigh *)> neigh_filter(
+      rtnl_neigh_alloc(), &rtnl_neigh_put);
+
+  rtnl_neigh_set_ifindex(neigh_filter.get(), ifindex);
+  rtnl_neigh_set_dst(neigh_filter.get(), addr);
+  rtnl_neigh_set_family(neigh_filter.get(), AF_INET);
+
+  nl_cache_foreach_filter(
+      nl->get_cache(cnetlink::NL_NEIGH_CACHE), OBJ_CAST(neigh_filter.get()),
+      [](struct nl_object *obj, void *arg) {
+        auto *add_list = static_cast<std::list<struct rtnl_neigh *> *>(arg);
+
+        add_list->emplace_back(NEIGH_CAST(obj));
+      },
+      neigh);
+
+  return 0;
+}
+
 // XXX separate function to make it possible to add lo addresses more directly
 int nl_l3::add_l3_addr(struct rtnl_addr *a) {
   assert(sw);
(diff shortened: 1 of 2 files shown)


src/netlink/nl_l3.cc
```diff
@@ -225,7 +245,17 @@ int nl_l3::add_l3_addr(struct rtnl_addr *a) {
...
+  if (std::list<struct rtnl_neigh *> neigh; search_neigh_cache(ifindex, addr, &neigh) == 0) {
+    VLOG(2) << __FUNCTION__ << ": list neigh size " << neigh.size();
+    if (neigh.size() > 0)
+      update = true;
+  }
+
+  if (prefixlen == 32)
+    rv = sw->l3_unicast_host_add(ipv4_dst, 0, false, update, vrf_id);
+  else
+    rv = sw->l3_unicast_route_add(ipv4_dst, mask, 0, false, update, vrf_id);
```
DFritzsche, line 258: If 'mask' is only used in the `sw->l3_unicast_route_add` function, I suggest to define mask with 0 and only set it when prefix == 32. This would remove code duplication here and on lines 240-243.

src/netlink/nl_l3.cc
```diff
@@ -225,7 +245,17 @@ int nl_l3::add_l3_addr(struct rtnl_addr *a) {
     return rv;
   }
 
-  rv = sw->l3_unicast_route_add(ipv4_dst, mask, 0, false, false, vrf_id);
+  bool update = false;
```
DFritzsche, line 248: 'update' variable can be defined earlier in the function and used instead of hard-coded 'false' in e.g. `sw->l3_unicast_route_add`

rubensfig: In https://github.com/bisdn/basebox/pull/301/files#diff-0ff22527108b0d3bbb8ff9b1050a2cdd341c2951e766c8e1e72f6a4412af9d25R243, the false value does have to be hardcoded, since we will not handle loopback address updates in this function. In this case, I think hardcoding makes it more explicit that we dont expect an update in this function.

src/netlink/nl_l3.cc
```diff
@@ -225,7 +245,17 @@ int nl_l3::add_l3_addr(struct rtnl_addr *a) {
...
+  if (std::list<struct rtnl_neigh *> neigh; search_neigh_cache(ifindex, addr, &neigh) == 0) {
+    VLOG(2) << __FUNCTION__ << ": list neigh size " << neigh.size();
+    if (neigh.size() > 0)
+      update = true;
+  }
+
+  if (prefixlen == 32)
+    rv = sw->l3_unicast_host_add(ipv4_dst, 0, false, update, vrf_id);
```
KanjiMonster, line 256: What happens if after ~5 minutes the neighbor times out and the kernel sends us a NEIGH_DEL message for this IP?

Won't this trigger a `del_l3_neigh`, which would then call `l3_unicast_host_remove(ipv4_dst)` and break everything?

Also what about IPv6? Or does VRRP only support IPv4?

src/netlink/nl_l3.cc
ideaship, line 258: How about adding a blank line after this one to make the code more readable?

ideaship:
I find the lack of comments in this changeset somewhat concerning. Yes, it largely matches the rest of the code, but still, shouldn't the code reader be given a hint on what is going on and why this is needed?

src/netlink/nl_l3.cc
```diff
@@ -813,16 +864,34 @@ int nl_l3::update_l3_neigh(struct rtnl_neigh *n_old, struct rtnl_neigh *n_new) {
...
   int family = rtnl_neigh_get_family(n);
+  bool skip_addr_remove = false;
 
-  assert(n);
   if (n == nullptr)
     return -EINVAL;
 
-  if (family == AF_INET) {
-    addr = rtnl_neigh_get_dst(n);
+  // delete l3 unicast group (mac rewrite)
```
KanjiMonster, line 878: You don't delete the l3 unicast group here, this comment should stay where it was

src/netlink/nl_l3.cc
```diff
@@ -813,16 +864,34 @@ int nl_l3::update_l3_neigh(struct rtnl_neigh *n_old, struct rtnl_neigh *n_new) {
...
+    if (struct nl_addr *link_addr = rtnl_addr_get_local(i);
+        nl_addr_cmp(addr, link_addr) == 0) {
+      VLOG(1) << __FUNCTION__
+              << ": address configured on the interface, skipping";
+      skip_addr_remove = true;
+      break;
+    }
+  }
+
+  if (family == AF_INET && skip_addr_remove) {
```
KanjiMonster, line 894: Did you mean `family == AF_INET && !skip_addr_remove`?

src/netlink/nl_l3.cc
```diff
@@ -813,16 +864,34 @@ int nl_l3::update_l3_neigh(struct rtnl_neigh *n_old, struct rtnl_neigh *n_new) {
 }
 
 int nl_l3::del_l3_neigh(struct rtnl_neigh *n) {
+  assert(n);
+
   int rv = 0;
-  struct nl_addr *addr;
+  struct nl_addr *addr = rtnl_neigh_get_dst(n);
+
```
KanjiMonster, line 871: unnecessary empty line

src/netlink/nl_l3.cc
```diff
@@ -831,9 +900,8 @@ int nl_l3::del_l3_neigh(struct rtnl_neigh *n) {
 
     // delete next hop
     rv = sw->l3_unicast_host_remove(ipv4_dst);
-  } else {
-    addr = rtnl_neigh_get_dst(n);
-
+  }
+  if (family == AF_INET6 && skip_addr_remove) {
```
KanjiMonster, line 904:
```suggestion
  } else if (family == AF_INET6 && skip_addr_remove) {
```

Also did you mean `!skip_addr_remove`?

src/netlink/nl_l3.cc
```diff
@@ -858,10 +926,6 @@ int nl_l3::del_l3_neigh(struct rtnl_neigh *n) {
     return rv;
   }
 
-  // delete l3 unicast group (mac rewrite)
-  int ifindex = rtnl_neigh_get_ifindex(n);
-  auto link = nl->get_link_by_ifindex(ifindex);
-
   if (!link)
     return -EINVAL;
```
KanjiMonster, line 930: At this point `link` can't be a nullptr anymore, you have already used it plenty. You need to move this check to top as well.

KanjiMonster (changes requested):

src/netlink/nl_l3.cc
```diff
@@ -813,16 +865,35 @@ int nl_l3::update_l3_neigh(struct rtnl_neigh *n_old, struct rtnl_neigh *n_new) {
...
+  if (!link)
+    return -EINVAL;
+
+  std::deque<rtnl_addr *> link_addresses;
+  get_l3_addrs(link.get(), &link_addresses);
+  for (auto i : link_addresses) {
+    if (struct nl_addr *link_addr = rtnl_addr_get_local(i);
+        nl_addr_cmp(addr, link_addr) == 0) {
+      VLOG(1) << __FUNCTION__
+              << ": address configured on the interface, skipping";
```
KanjiMonster, line 890: Which address, which interface? Either add the information, or drop the debug message. Especially since there is nothing else printing that information.

PR #303, opened 2021-05-05
cnetlink: delay thread running state until tap interfaces are created

rubensfig:
## Description
This PR aims to rework the netlink internal state transition from INIT to RUNNING. By setting the state to running after the tap interfaces are created, we delay baseboxd netlink objects processing until after the openflow channel is alive. 

## Motivation and Context
When running baseboxd with systemd-networkd configuring the network state, all configurations are usually set before the ofdpa agent is running. As such we risk losing route and address configuration notifications, thus leaving the switch in an unusable setup. This racy behaviour can be avoided by processing these notifications only after the openflow channel is established.

### DIFF
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -630,9 +630,11 @@ void cnetlink::nl_cb_v2(struct nl_cache *cache, struct nl_object *old_obj,
   assert(data);
   auto nl = static_cast<cnetlink *>(data);
 
-  // only enqueue nl msgs if not in stopped state
-  if (nl->state != NL_STATE_STOPPED)
-    nl->nl_objs.emplace_back(action, old_obj, new_obj);
+  // Enqueue the messages regardless of the state
+  // Due to the cache synch, we will enqueue the objects
+  // from the start of the netlink thread, ensuring baseboxd
+  // will process all events from the nl caches
+  nl->nl_objs.emplace_back(action, old_obj, new_obj);
 }
 
 void cnetlink::set_tapmanager(std::shared_ptr<tap_manager> tm) {
--- a/src/netlink/nbi_impl.h
+++ b/src/netlink/nbi_impl.h
@@ -25,6 +25,7 @@ public:
 
   // nbi
   void register_switch(switch_interface *) noexcept override;
+  void switch_state_notification(enum switch_state) noexcept override;
   void resend_state() noexcept override;
   void
   port_notification(std::deque<port_notification_data> &) noexcept override;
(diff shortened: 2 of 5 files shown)


src/netlink/cnetlink.cc
```diff
@@ -504,7 +504,6 @@ void cnetlink::handle_wakeup(rofl::cthread &thread) {
   switch (state) {
   case NL_STATE_INIT:
     init_subsystems();
```
KanjiMonster, line 506: `cnetlink::initsubsystems()` will call `nl_l3::init()` which will scan the lo interface for IP addresses and (try to) add flows for them, so delaying the RUNNING state is not enough.

KanjiMonster (changes requested):
The commit really needs a better description on what the issue is and how this is fixing it. 

The proper way to get this papered over is to revert https://github.com/bisdn/basebox/pull/257/files and remove https://github.com/bisdn/basebox/blob/master/src/netlink/cnetlink.cc#L633-L634 so we don't ignore netlink messages anymore before we got out of the _STOPPED state. 
Alternatively have an extra state for that so we can differentiate between we should not process them yet and we are shutting down/closing, so we won't process them ever.

But just not dropping them anymore is probably easiest. If baseboxd runs for so long it runs out of memory for storing these notifications, you will have bigger issues anyway.

src/netlink/nbi_impl.h
```diff
@@ -31,6 +31,7 @@ class nbi_impl : public nbi, public switch_callback {
   int enqueue(uint32_t port_id, basebox::packet *pkt) noexcept override;
   int fdb_timeout(uint32_t port_id, uint16_t vid,
                   const rofl::caddress_ll &mac) noexcept override;
+  void switch_state_notification(enum switch_state state) noexcept;
```
KanjiMonster, line 34: Why do you need this? This wasn't needed before, and your commit message provides no reasoning.

KanjiMonster (changes requested):
Please reorder your commits so you first enable unconditional enqueuing, and then revert the merge.

> This PR was reversed due to a change in how we process the initial state
of baseboxd. While the PR#257 was aiming to start the netlink
processing as early as possible, a bug was introduced where several
netlink messages will be processed imediately even though the switch
connection is not yet up. Reverting the commit restores the old behavior
to prepare for changed workflow.

This needs a better explanation that explicitly says what goes wrong. There are some types of configuration that we apply without checking if they are related to our interfaces, namely routes, and ip addresses on the lo interface.

Applying routes unconditionally becomes an issue if these routes are on vlan interfaces on bridge or bond interfaces, which may exist before basebox has an openflow connection and creates the port interfaces.

We should be ignoring routes on non-managed interfaces (i.e. bridges/bonds with no port interfaces attached), and only add them the moment a port interface gets attached to the bridge or bond.

But changing this requires larger changes in baseboxd, so delaying processing of netlink messages again works as a interim workaround.

And by changing the behavior of not dropping the netlink messages in stopped state anymore, we avoid re-introducing the issue that the original PR fixed.

Issue #305, opened 2021-06-04
baseboxd exception with GLOG_v=5 caused by initial netlink message


akoepsel:
## Expected Behavior
baseboxd should not crash ...

## Actual Behavior
An exception occurs when starting baseboxd with an increased logging verbosity level (GLOG_v >= 4), once the first netlink message is received from the underlying subsystems (kernel => netlink => libnl). The netlink message announces the loopback interface and causes the following exception:

```
I0604 14:07:31.449709    13 cnetlink.cc:578] handle_read_event: thread=cthread, tid: 139958419461888, num: 1, name: netlink 
timers list is EMPTY
, fd=6
I0604 14:07:31.449949    13 cnetlink.cc:626] : cache=0x564db0178170, diff=2 action=5 old_obj=0x564db01a78c0 new_obj=0x7f4a94034d40
I0604 14:07:31.450075    13 nl_obj.cc:77] incremented refcounts nl_obj=0x564db01799b0 (old_obj=0x564db01a78c0 new_obj=0x7f4a94034d40)
I0604 14:07:31.450120    13 nl_obj.cc:13] created nl_obj=0x564db01799b0 (old_obj=0x564db01a78c0 new_obj=0x7f4a94034d40)
I0604 14:07:31.450181    13 cnetlink.cc:582] handle_read_event: #processed=1
I0604 14:07:31.450318    13 nl_l3.cc:116] operator() : found configured loopback (nl_obj=0x564db01abd30) 127.0.0.1/8 inet dev lo scope host <permanent>
   label lo
   valid-lifetime forever preferred-lifetime forever
   created boot-time+1s 680msec last-updated boot-time+1s 680msec
I0604 14:07:31.450544    13 nl_bond.cc:35] get_lag_id: lag_id not found for if=1
--Type <RET> for more, q to quit, c to continue without paging--c

Thread 2 "netlink" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7f4a9be01700 (LWP 13)]
0x0000564dae125367 in std::char_traits<char>::length (__s=0x0) at /usr/include/c++/9/bits/char_traits.h:335
335             return __builtin_strlen(__s);
(gdb) bt
#0  0x0000564dae125367 in std::char_traits<char>::length (__s=0x0)
    at /usr/include/c++/9/bits/char_traits.h:335
[...]

PR #306, opened 2021-06-04
do not log netlink messages for loopback devices

akoepsel:
Signed-off-by: Andreas Koepsel <andreas.koepsel@bisdn.de>

## Description
Swapped code blocks on lines [69-71](https://github.com/bisdn/basebox/blob/master/src/netlink/netlink-utils.cc#L69) and [73-76](https://github.com/bisdn/basebox/blob/master/src/netlink/netlink-utils.cc#L73) in [src/netlink/netlink-utils.cc](https://github.com/bisdn/basebox/blob/master/src/netlink/netlink-utils.cc) to avoid an exception caused by a NULL (string) pointer when the set of initial netlink messages referring to the loopback device  is received and baseboxd's verbosity level causes the VLOG stream to call function rtnl_link_get_family(link).

## Motivation and Context
  - Solves issue #305

## How Has This Been Tested?
  - See issue #305 how to reproduce the exception
  - tested against master branch (exception occurs) and branches containing proposed changes (no exception occurs)

### DIFF
--- a/src/netlink/netlink-utils.cc
+++ b/src/netlink/netlink-utils.cc
@@ -66,14 +66,14 @@ enum link_type get_link_type(rtnl_link *link) noexcept {
     type = rtnl_link_get_type(link);
   }
 
-  VLOG(4) << __FUNCTION__ << ": type=" << std::string_view(type)
-          << ", slave=" << slave << ", af=" << rtnl_link_get_family(link)
-          << " of link " << OBJ_CAST(link);
-
   // lo has no type
   if (!type) {
     return LT_UNSUPPORTED;
   }
+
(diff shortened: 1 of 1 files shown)


KanjiMonster:
I don't think not logging is the right solution.

Maybe we can have a helper function something like
```
std::string_view safe_string_view(const char *str)
{
  if (str == nullptr)
    return std::string_view("(null)");
  else
    return std::string_view(str);
}
```

and then use this everywhere where we might have a null pointer? This would replicate `printf`'s behavior when passing NULL to a %s.

KanjiMonster:
And then squash all the issues in one (or two) commits.

rubensfig:
I have changed the target branch to next, to prepare for the merges.

jklare:
is this still being worked on?

Issue #307, opened 2021-06-04
baseboxd exception with GLOG_v=3 in cnetlink.cc on line 1080


akoepsel:
## Expected Behavior
baseboxd should not crash ...

## Actual Behavior
An exception occurs when starting baseboxd with an increased logging verbosity level (GLOG_v = 3 or above), once a netlink message calling method cnetlink::route_link_apply() is received causing the following exception: 

```
--Type <RET> for more, q to quit, c to continue without paging--

Thread 2 "netlink" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff6b28700 (LWP 13)]
0x00005555555ec367 in std::char_traits<char>::length (__s=0x0)
    at /usr/include/c++/9/bits/char_traits.h:335
335             return __builtin_strlen(__s);
(gdb) bt
#0  0x00005555555ec367 in std::char_traits<char>::length (__s=0x0)
    at /usr/include/c++/9/bits/char_traits.h:335
#1  0x000055555575ae84 in std::basic_string_view<char, std::char_traits<char> >::basic_string_view (this=0x7ffff6b27500, __str=0x0)
    at /usr/include/c++/9/string_view:124
#2  0x0000555555754da0 in basebox::cnetlink::link_updated (
    this=0x555555943970, old_link=0x7ffff00144c0, new_link=0x7ffff001b9f0)
    at ../src/netlink/cnetlink.cc:1083
#3  0x0000555555752ab0 in basebox::cnetlink::route_link_apply (
    this=0x555555943970, obj=...) at ../src/netlink/cnetlink.cc:831
#4  0x00005555557500f2 in basebox::cnetlink::handle_wakeup (
    this=0x555555943970, thread=...) at ../src/netlink/cnetlink.cc:531
#5  0x00007ffff7d13374 in rofl::cthread::handle_wakeup (this=0x555555943980)
    at /git/rofl-common/src/rofl/common/locking.hpp:47
#6  0x00007ffff7d14095 in rofl::cthread::run_loop (this=0x555555943980)
    at cthread.cpp:732
#7  0x00007ffff753c609 in start_thread (arg=<optimized out>)
    at pthread_create.c:477
#8  0x00007ffff7461293 in clone ()
    at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95

```

## Steps to Reproduce the Problem
  - Start baseboxd with GLOG_v=3.
  - Connect a switch
  - Wait some time for netlink events to occur ...

[...]

PR #308, opened 2021-06-04
do not call std::string_view() with NULL pointer as argument

akoepsel:
## Description

In [cnetlink.cc line 1075](https://github.com/bisdn/basebox/blob/master/src/netlink/cnetlink.cc#L1075) the VLOG call must not call std::string_view(NULL). 
Added two checks to prevent a situation where calling rtnl_link_get_type() or rtnl_link_get_slave_type() yield a NULL pointer, inevitably causing an exception.

## Motivation and Context

  - Solves issue #307 

## How Has This Been Tested?

  - See issue #307 how to reproduce the exception
  - tested against master branch (exception occurs) and branch containing proposed changes (no exception occurs)

### DIFF
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -1072,15 +1072,19 @@ void cnetlink::link_updated(rtnl_link *old_link, rtnl_link *new_link) noexcept {
   int af_old = rtnl_link_get_family(old_link);
   int af_new = rtnl_link_get_family(new_link);
 
-  VLOG(3) << __FUNCTION__ << ": old_link_type="
-          << std::string_view(rtnl_link_get_type(old_link))
-          << ", new_link_type="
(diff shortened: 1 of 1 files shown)


rubensfig:
I have changed the branch to next, to prepare for the merges.

@KanjiMonster could you please add your comment from #306 here as well?

KanjiMonster:
I don't think not logging is the right solution.

Maybe we can have a helper function something like
```
std::string_view safe_string_view(const char *str)
{
  if (str == nullptr)
    return std::string_view("(null)");
  else
    return std::string_view(str);
}
```

and then use this everywhere where we might have a null pointer? This would replicate `printf`'s behavior when passing NULL to a %s.

Finally, let's maybe have only one PR for this.

jklare:
is this still being worked on?

PR #315, opened 2021-06-08
nl_l3: don't add fe80::/10 for IPv6 link local addresses

rubensfig:
While the address space reserved for link local addresses is indeed
fe80::/10, the actual used network is fe80::/64. All link local
addresses assigned by the kernel or e.g. systemd follow that, so we can
just treat them like any other network address.

## Motivation and Context

The onlink route is a /64, creating a /10 is wrong.

## How Has This Been Tested?

Testing pipeline run on Questone 2A with adapted flow checks.


KanjiMonster:
Rebased onto latest next.

PR #323, opened 2021-06-25
Handle MAC address changes on tap interfaces

rubensfig:
## Description
This PR adds handling of changing MAC addresses on the TAP interfaces. When doing commands like,

```
$ ip link set PORT address MAC_ADDRESS 
```

The linux mac address will change to the one specified. When this change is triggered, kernel sends two notifications, the first one related to the change of the address itself. The next one will be flushing the ip neighs on the interface. 

The changing of mac addresses on the tap interfaces is relevant when there are IP addresses/ neighs/ routes configured on the interface. baseboxd must pick up the notification and change the entries in the termination mac table and egress l3 unicast group. Since there are maps following the entries in these tables, and associated refcounts, they must also be updated.

## Motivation and Context
Setting new mac addresses in tap interfaces is a management operation that should be supported in baseboxd. After the mac are reset, connectivity should still continue. For example, since https://github.com/systemd/systemd/commit/550c8784c5, the mac addresses will be reset on baseboxd startup. If there are races between baseboxd setting the ASIC entries and the rewrite of the mac address, that can potentially leave pending entries in the asic, making the system unusable.

### DIFF
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -551,6 +551,10 @@ int cnetlink::remove_l3_configuration(rtnl_link *link) {
   return rv;
 }
 
+int cnetlink::update_on_mac_change(rtnl_link *old_link, rtnl_link *new_link) {
+  return 0;
+}
+
 bool cnetlink::has_l3_addresses(rtnl_link *link) {
   assert(link);
 
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -552,7 +552,33 @@ int cnetlink::remove_l3_configuration(rtnl_link *link) {
 }
 
 int cnetlink::update_on_mac_change(rtnl_link *old_link, rtnl_link *new_link) {
-  return 0;
+  int rv = 0;
+  int port_id = get_port_id(old_link);
+  uint16_t vid = vlan->get_vid(old_link);
+  struct nl_addr *old_mac = rtnl_link_get_addr(old_link);
+  struct nl_addr *new_mac = rtnl_link_get_addr(new_link);
+
+  rv = l3->update_l3_termination(port_id, vid, old_mac, new_mac);
+  if (rv < 0)
+    VLOG(1) << __FUNCTION__ << ": failed to update termination MAC, old link="
+            << OBJ_CAST(old_link) << " new link=" << OBJ_CAST(new_link);
+
+  // In response to the MAC address change on the interface, linux deletes the
+  // neighbors configured on the interface. We are tracking the state
+  // of the nexthops in the l3_interface_mapping in the nl_l3 class, by storing
+  // the port id, vid, src mac and dst mac.  The neighbor deletion messages are
+  // received for all neighbors on the link.
+  // While these messages will delete the link neighbors after the MAC address
+  // changes, they do not carry the source mac, which forces a lookup for it
+  // returning the updated one. We need to make sure the deletions find their
+  // targets by updating the entry and the map.
+  rv = l3->update_l3_egress(port_id, vid, old_mac, new_mac);
+  if (rv < 0)
+    VLOG(1) << __FUNCTION__
+            << ": failed to update l3 egress, old link=" << OBJ_CAST(old_link)
+            << " new link=" << OBJ_CAST(new_link);
+
+  return rv;
 }
 
 bool cnetlink::has_l3_addresses(rtnl_link *link) {
@@ -1299,6 +1325,14 @@ void cnetlink::link_updated(rtnl_link *old_link, rtnl_link *new_link) noexcept {
           << ", new_link_slave_type="
           << std::string_view(rtnl_link_get_slave_type(new_link))
           << ", af_old=" << af_old << ", af_new=" << af_new;
+
+  if (nl_addr_cmp(rtnl_link_get_addr(old_link), rtnl_link_get_addr(new_link)) &&
+      is_switch_interface(old_link)) {
+    if (update_on_mac_change(old_link, new_link) < 0)
+      LOG(ERROR) << __FUNCTION__
+                 << ": failed to update MAC old_link=" << OBJ_CAST(old_link)
+                 << " new_link=" << OBJ_CAST(new_link);
+  }
 
   if (af_old != af_new) {
     VLOG(1) << __FUNCTION__ << ": af changed from " << af_old << " to "
--- a/src/netlink/nl_l3.cc
+++ b/src/netlink/nl_l3.cc
@@ -476,7 +476,7 @@ int nl_l3::del_l3_addr(struct rtnl_addr *a) {
   rofl::caddress_ll mac = libnl_lladdr_2_rofl(addr);
 
   rv = del_l3_termination(port_id, vid, mac, family);
-  if (rv < 0) {
+  if (rv < 0 && rv != -ENODATA) {
     LOG(ERROR) << __FUNCTION__
                << ": failed to remove l3 termination mac(local) vid=" << vid
                << "; rv=" << rv;
@@ -1200,7 +1200,7 @@ int nl_l3::del_l3_termination(uint32_t port_id, uint16_t vid,
         << __FUNCTION__
         << ": tried to delete a non existing termination mac for port_id="
         << port_id << ", vid=" << vid << ", mac=" << mac << ", af=" << af;
-    return 0;
+    return -ENODATA;
   }
 
   // still existing references
--- a/src/netlink/nl_l3.cc
+++ b/src/netlink/nl_l3.cc
@@ -68,7 +68,7 @@ std::unordered_map<
     l3_interface>
     l3_interface_mapping;
 
-// key: source port_id, vid, src_mac, ethertype ; value: refcount
+// key: source port_id, vid, src_mac, af ; value: refcount
 std::unordered_multimap<std::tuple<int, uint16_t, rofl::caddress_ll, uint16_t>,
                         int>
     termination_mac_mapping;
--- a/src/netlink/nl_l3.cc
+++ b/src/netlink/nl_l3.cc
@@ -641,7 +641,9 @@ int nl_l3::del_l3_neigh_egress(struct rtnl_neigh *n) {
   // remove egress group reference
   int rv = del_l3_egress(ifindex, vid, s_mac, d_mac);
   if (rv < 0 and rv != -EEXIST) {
-    LOG(ERROR) << __FUNCTION__ << ": failed to add l3 egress";
+    LOG(ERROR) << __FUNCTION__
+               << ": failed to remove l3 egress ifindex=" << ifindex
+               << " vid=" << vid;
   }
 
   return rv;
@@ -1142,6 +1144,9 @@ int nl_l3::add_l3_termination(uint32_t port_id, uint16_t vid,
     if (i->first == needle) {
       // found, increment refcount
       i->second++;
+      VLOG(4) << __FUNCTION__ << ": incremented refcount=" << i->second
+              << " for key portid=" << port_id << " vid=" << vid
+              << " mac=" << mac << " af=" << af;
       return 0;
     }
   }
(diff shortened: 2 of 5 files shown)


src/netlink/nl_interface.cc
```diff
@@ -17,6 +17,11 @@ int nl_interface::changed(rtnl_link *old_link, rtnl_link *new_link) noexcept {
     tm->update_mtu(new_link);
   }
 
+  if (nl_addr_cmp(rtnl_link_get_addr(old_link), rtnl_link_get_addr(new_link))) {
```
KanjiMonster, line 20: Why not move this part to `cnetlink::link_updated()` and call ``l3->update_l3_termination(old_link, new_link)`` directly, instead of going through here? Then you can move it outside of the switch case and don't need to explicitly handle the interface type (vlan,tap,bond,bridge,..)

src/netlink/nl_interface.cc
```diff
@@ -17,6 +17,11 @@ int nl_interface::changed(rtnl_link *old_link, rtnl_link *new_link) noexcept {
...
+    nl->update_on_mac_change(old_link, new_link);
+    return 0;
```
KanjiMonster, line 22: Redundant return call here (and might not be wanted at all).

KanjiMonster (changes requested):

src/netlink/nl_interface.cc
rubensfig: This function is already an handler for picking changes on interfaces. I would suggest leaving this here for code readability and making everything concise, instead of spreading the changes all over the place.

src/netlink/nl_l3.cc
```diff
@@ -1228,6 +1233,161 @@ int nl_l3::del_l3_termination(uint32_t port_id, uint16_t vid,
...
+      std::begin(l3_interface_mapping), std::end(l3_interface_mapping),
+      [needle, &update_l3](
+          std::pair<
+              std::tuple<int, uint16_t, rofl::caddress_ll, rofl::caddress_ll>,
+              l3_interface>
+              l3_iface_tuple) {
+        if (std::get<1>(l3_iface_tuple.first) == std::get<1>(needle) &&
+            std::get<2>(l3_iface_tuple.first) == std::get<2>(needle))
+          update_l3.emplace(l3_iface_tuple.first, l3_iface_tuple.second);
+      });
```
KanjiMonster, line 1358: Any reason you can't just do
```
  for (it : l3_interface_mapping) {
    if (std::get<1>it.first == vid && std::get<2>it.first == old_mac)
      update_l3.emplace(it.first, it.second);
  }
```

KanjiMonster (changes requested):
I don't see you handling mac address changes on bond interfaces directly.

Also I'm wondering about the worth of the iface class, and if it wouldn't make more sense to do the calls/checks directly in `cnetlink::link_updated()`.

src/netlink/nl_l3.cc
```diff
@@ -632,16 +632,19 @@ int nl_l3::del_l3_neigh_egress(struct rtnl_neigh *n) {
   if (nl->is_bridge_interface(ifindex)) {
     auto fdb_res = nl->search_fdb(vid, d_mac);
 
-    assert(fdb_res.size() == 1);
-    ifindex = rtnl_neigh_get_ifindex(fdb_res.front());
+    // assert(fdb_res.size() == 1);
+    if (fdb_res.size() == 1)
+    	ifindex = rtnl_neigh_get_ifindex(fdb_res.front());
+    else if (fdb_res.size() == 0)
+	 VLOG(1) << __FUNCTION__ << ": fdb_res = 0";
```
KanjiMonster: This change seems to be unrelated?

src/netlink/nl_l3.cc
```diff
@@ -1228,6 +1230,148 @@ int nl_l3::del_l3_termination(uint32_t port_id, uint16_t vid,
...
+    VLOG(4) << __FUNCTION__
+            << ": no mapping found, no egress entry to be updated";
+    return 0;
+  }
+
+  // update the switch egress entry with the new mac address
+  for (auto it : update_l3) {
+    uint32_t l3_iface_id = it.second.l3_interface_id;
+    if (port_id == 0)
+      port_id = std::get<0>(it.first);
```
KanjiMonster, line 1354: might as well always use the port_id from it.first, it should never be different unless passed on port_id is 0.

src/netlink/nl_l3.cc
```diff
@@ -1228,6 +1230,148 @@ int nl_l3::del_l3_termination(uint32_t port_id, uint16_t vid,
...
+      std::tuple<int, uint16_t, rofl::caddress_ll, rofl::caddress_ll>,
+      l3_interface>
+      update_l3;
+
+  // vlan, mac address combination is unique, get the entries that match
+  // not parsing for the port id solves the issue for bridge interfaces
+  // that portid is 0 and the entries are gotten from the fdb
+  for (auto it : l3_interface_mapping) {
+    if (std::get<1>(it.first) == vid &&
+        std::get<2>(it.first) == libnl_lladdr_2_rofl(old_mac))
```
KanjiMonster, line 1340: maybe calculate `libnl_lladdr_2_rofl(old_mac)` once at the beginning instead of doing it at each comparison?

src/netlink/nl_l3.cc
```diff
@@ -1228,6 +1230,148 @@ int nl_l3::del_l3_termination(uint32_t port_id, uint16_t vid,
...
+  // update the switch egress entry with the new mac address
+  for (auto it : update_l3) {
+    uint32_t l3_iface_id = it.second.l3_interface_id;
+    if (port_id == 0)
+      port_id = std::get<0>(it.first);
+    rv = sw->l3_egress_update(port_id, vid, libnl_lladdr_2_rofl(new_mac),
+                              std::get<3>(it.first), &l3_iface_id);
+
+    auto l3_if_new_tuple = std::make_tuple(
+        port_id, vid, libnl_lladdr_2_rofl(new_mac), std::get<3>(it.first));
```
KanjiMonster, line 1359: likewise with `libnl_lladdr_2_rofl(new_mac)`, calculate it once at the beginning, then reuse it.

KanjiMonster (changes requested):
Please squash fixes into their appropriate commits, and move any formatting fixes to the start, not in the middle.

Also I still don't see bond interface mac address changes being addressed.

src/netlink/cnetlink.cc
```diff
@@ -1317,6 +1322,7 @@ void cnetlink::link_updated(rtnl_link *old_link, rtnl_link *new_link) noexcept {
   case LT_VLAN: {
     if (lt_new != LT_VRF_SLAVE) {
       VLOG(1) << __FUNCTION__ << ": ignoring update of VLAN interface";
+      iface->changed(old_link, new_link);
```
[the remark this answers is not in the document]
KanjiMonster: Is this still needed? You are probably now calling this twice for LT_VLAN I think?

src/netlink/cnetlink.cc
```diff
@@ -1287,6 +1309,9 @@ void cnetlink::link_updated(rtnl_link *old_link, rtnl_link *new_link) noexcept {
           << std::string_view(rtnl_link_get_slave_type(new_link))
           << ", af_old=" << af_old << ", af_new=" << af_new;
 
+  if (nl_addr_cmp(rtnl_link_get_addr(old_link), rtnl_link_get_addr(new_link)))
+    update_on_mac_change(old_link, new_link);
```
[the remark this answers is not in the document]
KanjiMonster: Also you might want to make sure that old_link is a device we care about, else you might pick up mac changes on foreign devices (like the management interface) which will all be port_id 0.

src/netlink/nl_l3.cc
```diff
@@ -1144,6 +1144,7 @@ int nl_l3::add_l3_termination(uint32_t port_id, uint16_t vid,
     if (i->first == needle) {
       // found, increment refcount
       i->second++;
+      VLOG(4) << __FUNCTION__ << ": found new refcount=" << i->second;
```
KanjiMonster, line 1147: This output doesn't help much if you don't print for what termination mac entry it is.

KanjiMonster (changes requested):

src/netlink/cnetlink.cc
```diff
@@ -551,6 +551,35 @@ int cnetlink::remove_l3_configuration(rtnl_link *link) {
...
+            << OBJ_CAST(old_link) << " new link=" << OBJ_CAST(new_link);
+
+  // In response to the MAC address change on the interface, linux deletes the
+  // neighbors configured on the interface. We are tracking the state
+  // of the nexthops in the l3_interface_mapping in the nl_l3 class, by storing
+  // the port id, vid, src mac and dst mac. The neighbor deletion messages
+  // are received for all neighbors on the link. While these messages will
+  // delete the link neighbors before the MAC address changes, they will carry
+  // the updated source MAC address. We need to make sure the deletions find
+  // their targets by updating the entry and the map.
```
KanjiMonster, line 573: > they will carry the updated source MAC address.

They don't carry any source mac address at all, which is why we need to look it up, and only see the updated mac address. If they had a source mac address, we wouldn't have that issue.

The messages we get are

1. link Foo changed mac from A to B
2. neigbhor with IP 1 and Mac X was deleted on link Foo

So at step 2, we do not have access to the old mac (foo), and we need to ask someone (nl cache), and they will reply with the current, updated mac.

Btw, you can use paragraphs, they help readability.

src/netlink/nl_l3.cc
```diff
@@ -1142,6 +1144,9 @@ int nl_l3::add_l3_termination(uint32_t port_id, uint16_t vid,
     if (i->first == needle) {
       // found, increment refcount
       i->second++;
+      VLOG(4) << __FUNCTION__ << ": found new refcount=" << i->second
```
KanjiMonster, line 1147:
```suggestion
      VLOG(4) << __FUNCTION__ << ": incremented refcount=" << i->second
```

src/netlink/nl_l3.cc
```diff
@@ -1185,7 +1193,9 @@ int nl_l3::del_l3_termination(uint32_t port_id, uint16_t vid,
     if (i->first == needle) {
       // found, decrement refcount
       --i->second;
-      VLOG(4) << __FUNCTION__ << ": found new refcount=" << i->second;
+      VLOG(4) << __FUNCTION__ << ": found new refcount=" << i->second
+              << " for key portid=" << port_id << " vid=" << vid
+              << " mac=" << mac << " af=" << af;
```
KanjiMonster, line 1198:
```suggestion
      VLOG(4) << __FUNCTION__ << ": decremented refcount=" << i->second
              << " for l3 terminaton for portid=" << port_id << " vid=" << vid
              << " mac=" << mac << " af=" << af;
```

src/netlink/nl_l3.cc
```diff
@@ -1228,6 +1238,129 @@ int nl_l3::del_l3_termination(uint32_t port_id, uint16_t vid,
...
+    if (rv < 0) {
+      VLOG(3) << __FUNCTION__
+              << ": failed to add termination mac port=" << port_id
+              << " vid=" << vid << " mac=" << n_mac;
+      return rv;
+    }
+
+    auto new_key = std::make_tuple(port_id, vid, n_mac, AF_INET);
+    termination_mac_mapping.emplace(std::move(new_key), refcnt);
+    termination_mac_mapping.emplace(std::move(new_key), refcnt);
```
KanjiMonster, line 1276: why twice?

KanjiMonster (changes requested):

src/netlink/cnetlink.cc
```diff
@@ -551,6 +551,36 @@ int cnetlink::remove_l3_configuration(rtnl_link *link) {
...
+
+  // In response to the MAC address change on the interface, linux deletes the
+  // neighbors configured on the interface. We are tracking the state
+  // of the nexthops in the l3_interface_mapping in the nl_l3 class, by storing
+  // the port id, vid, src mac and dst mac.  The neighbor deletion messages are
+  // received for all neighbors on the link.
+  // While these messages will delete the link neighbors before the MAC address
+  // changes, they do not carry the source mac, which forces a lookup for it
+  // returning the updated one. We need to make sure the deletions find their
+  // targets by updating the entry and the map.
```
KanjiMonster, line 574: The deletions come after the source (interface) mac changed, else we wouldn't have the problem.

KanjiMonster (changes requested):
Please also squash in fixes/suggestions into their respective commits

KanjiMonster (approved):

PR #332, opened 2021-08-19
Adapt libnl bridge vlan usage to updated API

KanjiMonster:
With https://github.com/bisdn/meta-switch/pull/2 we simplify the libnl bridge vlan API. Since this changes the API, we need to update the code here as well.

## How Has This Been Tested?

Ran on tonights pipelines.

Depends on https://github.com/bisdn/meta-switch/pull/2

### DIFF
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -1699,15 +1699,16 @@ void cnetlink::route_mdb_apply(const nl_obj &obj) {
 }
 
 void cnetlink::route_bridge_vlan_apply(const nl_obj &obj) {
-  assert(obj.get_new_obj());
 
-  switch (obj.get_msg_type()) {
-  case RTM_NEWVLAN:
+  switch (obj.get_action()) {
+  case NL_ACT_NEW:
+  case NL_ACT_CHANGE:
+    assert(obj.get_new_obj());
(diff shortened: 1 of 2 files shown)


src/netlink/cnetlink.cc
```diff
@@ -1699,19 +1699,17 @@ void cnetlink::route_mdb_apply(const nl_obj &obj) {
...
       bridge->set_pvlan_stp(BRIDGE_VLAN_CAST(obj.get_new_obj()));
     break;
-#if 0
-  case RTM_DELVLAN:
-	if (bridge)
-	        bridge->set_pvlan_stp(BRIDGE_VLAN_CAST(obj.get_new_obj()));
-	break;
-#endif
+  case NL_ACT_DEL:
+    VLOG(1) << __FUNCTION__ << ": removing vlan stg membership not supported";
```
rubensfig, line 1711: Is VLOG(1) the better level for this? Wouldnt log info or warning be more explicit and avoid confusions?

KanjiMonster: We have other "not implemented" messages at the same level. This isn't something the user actively sets, so printing a INFO or WARNING level message is probably only confusing, especially since this isn't something the user can do anything about it.

PR #333, opened 2021-08-23
add proper lifecycle management of vlan STGs

KanjiMonster:
Add proper lifecycle management for STP STGs so they get deleted once not needed anymore.

## Description
Currently STP STGs are only created, but never destroyed any more. To fix this, add the required implementations for removing STGs, and implement handling bridge vlan removal notifications. Since we need to know when we can remove the group as well, let our internal state cache for calculation the effective STP states notify if it creates the map for a vlan (i.e. we need to create the STG), and if it removes the map because it removed the last member (i.e. we can drop the STG).

Since this will eventually lead to us needing to reuse STG IDs, add code for finding the next free ID, and keep track of used IDs in a bitmap.

## How Has This Been Tested?

Ran a full pipeline on the Questone 2As (17666)


src/of-dpa/controller.cc
```diff
@@ -2186,6 +2186,21 @@ int controller::lookup_stpid(uint32_t vlan_id) noexcept {
...
+int controller::find_free_stgid() noexcept {
+  // 512 groups, 2 reserved (0 and 1)
+  if (vlan_to_stg.size() == (512 - 2))
+    return -ENOSPC;
+
+  while (stg_in_use[current_stg]) {
+    current_stg++;
+
+    if (current_stg == 512)
+      current_stg = 2;
```
hilmarm, line 2198: If we use all the STGs, will this end up in an infinite loop, or do we make sure somewhere to not call this function if all STGs are in use?

KanjiMonster: A few lines above I check if vlan_to_stg has the maximum amount of supported entries (512 - 2 == 510), and return with -ENOSPC in that case. So if we reach that loop, we must have less than that, and there must be a free spot.

KanjiMonster:
Haven't done a style check, so marking temporarily as draft

KanjiMonster:
Fixed all one style issue.

hilmarm:
Seems to work fine for adding and deleting stg groups, tested on celestica questone.

How is the user notified about too many STG groups being created? i.e. what should happen if you try to add more than 512 STGs?

KanjiMonster:
>  i.e. what should happen if you try to add more than 512 STGs?

Basebox will log an error message failing to create more STG. There isn't really we can do much about it, as we cannot prevent the user from creating more than 510 different vlans on a bridge.

hilmarm:
> > i.e. what should happen if you try to add more than 512 STGs?
> 
> Basebox will log an error message failing to create more STG. There isn't really we can do much about it, as we cannot prevent the user from creating more than 510 different vlans on a bridge.

This is the basebox log when creating more than 512VLANs:

```get_members: lag does not exist for port53 ether ba:07:e3:9f:30:98 master swbridge <broadcast,multicast,up,running,lowerup> group 0```

Should we add an additional log warning when this happens?

KanjiMonster:
> > > i.e. what should happen if you try to add more than 512 STGs?
> > 
> > 
> > Basebox will log an error message failing to create more STG. There isn't really we can do much about it, as we cannot prevent the user from creating more than 510 different vlans on a bridge.
> 
> This is the basebox log when creating more than 512VLANs:
> 
> `get_members: lag does not exist for port53 ether ba:07:e3:9f:30:98 master swbridge <broadcast,multicast,up,running,lowerup> group 0`

This is a completely unrelated message.

> 
> Should we add an additional log warning when this happens?

Added an error log if this happens.

PR #340, opened 2021-09-01
fix various issues in multicast code and unbreak IPv6 snooping

KanjiMonster:
Fix various issues in multicast handling to follow specs and allow IPv6.

## Description

Fixes the following issues in the multicast code:

* fix debug messages on removal
* align add and remove code to do the same checks
* do not abort on the first ignored entry (go through them all)
* do not create/remove groups for 224.0.0.0/8, these should always be flooded
* do not ignore ff02::/10, but only ff::/15 and ff02::1, allowing IPv6 snooping to actually work

## How Has This Been Tested?

Running a pipeline and updating the multicast-ipv4 test to handle the additional IPv6 groups created.


rubensfig:
LGTM

rubensfig (approved):

PR #342, opened 2021-09-13
controller: l3_egress_remove: fix race with dropping references

KanjiMonster:
When l3 neighbors that are also route nexthop become unavailable, we
will send in rapid succession an update for the route to point to
controller instead of the l3 unicast group, and the removal of the l3
unicast group.

If the removal is processed before the route update, the removal will
fail since it still is referenced by the route, and the l3 interface ids
in baseboxd and ofdpa become out of sync, causing various issues.

Fix this by adding a barrier before trying to remove the l3 unicast
group to ensure that any pending route updates finished processing.


KanjiMonster:
Without this, it the rip-ipv4 test fails within a few rounds on the as4610s, while with it it didn't fail running for about two hours in a loop.

hilmarm:
lgtm

PR #348, opened 2021-10-29
handle neighbors moving between ports

KanjiMonster:
Properly update forwarding entries for known neighbors moving to new ports.

Depends on OF-DPA changes actually notifying us about station moves.

## Description

Handle moving stations between ports by properly detecting if this happens, and updating the ext_learn entry in the kernel and the flow entry. 

## Motivation and Context

Moving between ports may happen if e.g. people move their laptops, or if roaming from one AP to another. Not updating the entry means no connectivity until the old entry times out.

## How Has This Been Tested?

Tested on as4610 (and questone-2a with slightly different port names)

Switch 1:

```
ip link add name swbridge type bridge vlan_filtering 1
ip link set port2 master swbridge
ip link set port3 master swbridge
ip link set port54 master swbridge
ip link set swbridge up
ip link set port2 up
ip link set port3 up
ip link set port54 up
```

Switch 2:

```
ip link set port54 up
ip address add 10.0.10.1/24 dev port54
```

Server 1:

```
ip link add name swbridge type bridge vlan_filtering 1
ip link set swbridge address '02:00:00:00:00:01'
ip address add 10.0.10.2/24 dev swbridge
ip link set up swbridge
```

then either:

```
ip link set eno2 up
ip link set eno3 up

ip link set eno2 master swbridge
ping 10.0.10.1 -c 3
ip link set eno2 nomaster
ip link set eno3 master swbridge
ping 10.0.10.1 -c 3
```

or 

```
ip link set eno2 master swbridge
ip link set eno3 master swbridge

ip link set eno2 up
ping 10.0.10.1 -c 3
ip link set eno2 down
ip link set eno3 up
ping 10.0.10.1 -c 3
```

### DIFF
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -1559,10 +1559,25 @@ void cnetlink::neigh_ll_created(rtnl_neigh *neigh) noexcept {
   }
 }
 
-void cnetlink::neigh_ll_updated(__attribute__((unused)) rtnl_neigh *old_neigh,
-                                __attribute__((unused))
+void cnetlink::neigh_ll_updated(rtnl_neigh *old_neigh,
                                 rtnl_neigh *new_neigh) noexcept {
-  LOG(WARNING) << __FUNCTION__ << ": neighbor update not supported";
+  if (!check_ll_neigh(old_neigh) || !check_ll_neigh(new_neigh))
+    return;
+
+  int old_ifindex = rtnl_neigh_get_ifindex(old_neigh);
+  rtnl_link *old_base_link =
+      get_link(old_ifindex, AF_UNSPEC); // either tap, vxlan, ...
+  int new_ifindex = rtnl_neigh_get_ifindex(new_neigh);
+  rtnl_link *new_base_link =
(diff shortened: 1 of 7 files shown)


ideaship:
I wonder how that test could ever work considering that '02:00:00:00:01' is not a valid MAC address.

KanjiMonster:
> I wonder how that test could ever work considering that '02:00:00:00:01' is not a valid MAC address.

Because I typo'd when writing down what I did and missed a '00:'. Should be obviously 6 octets, fixed it.

src/netlink/cnetlink.cc
```diff
@@ -1562,7 +1562,19 @@ void cnetlink::neigh_ll_created(rtnl_neigh *neigh) noexcept {
 void cnetlink::neigh_ll_updated(__attribute__((unused)) rtnl_neigh *old_neigh,
                                 __attribute__((unused))
```
rubensfig, line 1563: Can we drop the unused atttribute now?

KanjiMonster: Yes, we even should, good catch.

Updated the commits, and also fixed all style issues I found.

rubensfig (approved):

PR #349, opened 2021-11-01
nl_l3: fix check for address family

rubensfig:
The check for avoiding adding the ::1/128 address is broken due to a
false AF.

Signed-off-by: Rubens Figueiredo <rubens.figueiredo@bisdn.de>


hilmarm:
lgtm, i added a test for this as well.

KanjiMonster (approved):
Good catch!

PR #353, opened 2021-11-16
simplify layer 2 multicast SAI interface 

KanjiMonster:
Sinplify the SAI multicast interface into just join and leave calls, and hide the implementation details in controller.cc.

## Description
Currently l2 multicast handling is split into two sets functions, one for handling the multicast group and flow, and one for handling the joined ports to that group.

Since baseboxd does not track the number of group listeners, it needs the functions to tell it when to add or remove the group and flow. This makes it a bit awkward to use, so instead of forcing baseboxd to handle that, let the controller implementaton do that and just expose "generic" join and leave calls that do the flow/group creation in the background.

## Motivation and Context
It makes the code a bit more clearer and understandable.

## How Has This Been Tested?
Running multicast-ipv{4,6} on questones.


rubensfig (approved):
LGTM, good refactoring of the code

PR #355, opened 2021-11-19
improve bridge ll neighbor handling

KanjiMonster:
Make sure we do not try to handle l2 neighbors for other bridges, and be silent about them.

## Description
We may have additional bridges on the system, like `docker0`. So getting notifications about changes on a bridge without it being "our" bridge is perfectly normal.

## Motivation and Context
Appearance of the docker bridge might cause issues due to not properly filtering the notifications.

## How Has This Been Tested?
Weekend pipeline on the AS4610.

### DIFF
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -1555,8 +1555,7 @@ void cnetlink::neigh_ll_created(rtnl_neigh *neigh) noexcept {
     // XXX TODO maybe we have to do this as well wrt. local bridging
     // normal bridging
     try {
-      if (bridge)
-        bridge->add_neigh_to_fdb(neigh);
+      bridge->add_neigh_to_fdb(neigh);
     } catch (std::exception &e) {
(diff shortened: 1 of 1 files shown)


src/netlink/cnetlink.cc
```diff
@@ -1502,6 +1502,11 @@ bool cnetlink::check_ll_neigh(rtnl_neigh *neigh) noexcept {
     return false;
   }
 
+  if (rtnl_neigh_get_master(neigh) != bridge->get_ifindex()) {
+    VLOG(1) << __FUNCTION__ << ": neighbor not on our bridge";
```
rubensfig, line 1506: Can we have a log of the specific neighbor we are trying to add?

KanjiMonster: Reworded it slightly differently so that neigh comes last like it does in the debug print after that.

rubensfig (approved):

PR #357, opened 2021-11-24
fix applying per vlan STP state to bond interfaces

KanjiMonster:
Fix applying per vlan STP state for bond interfaces by applying the state to all members of the bond group.

## Description
STP state is a state of physical ports, not logical ports in the switch, so when the kernel changes the stp state of a bond interface of a bridge, we need to propagate the state to all members of the bond interface.

We also need to make sure that when adding or removing an interface to a bond group that is also part of a bridge that we apply the currently configured stp state to the new bond member, or deconfigure it when the interface leaves.

## Motivation and Context
Allows using (M)STP on bridges with bond interfaces together.

## How Has This Been Tested?
Created a bridge with two bonds between both as4610 in testing (using ports 5/10 and 15/20), and checking the resulting stg stp states.

### DIFF
--- a/src/netlink/nl_bridge.cc
+++ b/src/netlink/nl_bridge.cc
@@ -1242,4 +1242,16 @@ int nl_bridge::drop_pvlan_stp(struct rtnl_bridge_vlan *bvlan_info) {
   return err;
 }
 
+std::map<uint16_t, uint8_t>
+nl_bridge::get_port_vlan_stp_states(rtnl_link *link) {
+  if (get_stp_state() == STP_STATE_DISABLED)
+    return {};
+
+  uint32_t port_id = nl->get_port_id(link);
+  if (port_id == 0)
+    return {};
+
+  return bridge_stp_states.get_min_states(port_id);
+}
+
 } /* namespace basebox */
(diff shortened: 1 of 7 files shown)


src/netlink/nl_bridge.cc
```diff
@@ -1127,23 +1127,33 @@ int nl_bridge::mdb_entry_remove(rtnl_mdb *mdb_entry) {
...
 
-  auto state = stp_state_to_string(stp_state);
   if (nbi::get_port_type(port_id) == nbi::port_type_lag) {
     auto members = nl->get_bond_members_by_port_id(port_id);
+    auto err2 = 0;
     for (auto mem : members) {
-      sw->ofdpa_stg_state_port_set(mem, 0, state);
+      err2 = sw->ofdpa_stg_state_port_set(mem, vid, state);
+      if (err2 < 0)
+        err = err2;
```
hilmarm, line 1140: so err2 is simply the error variable for each member of the bond?

KanjiMonster: This is to make sure that `err` contains the last error code, and not the return value of the last call to `ofdpa_stg_state_port_set`. So that if any call fails, we return an error code, and not only if the last one fails.

hilmarm: thanks, now it all makes sense

hilmarm:
Going through the code everything looks good to me, I will just need to try some testing on a physical switch

rubensfig (approved):

PR #359, opened 2021-12-08
nbi_impl: set link state last for tap devices

KanjiMonster:
Software or drivers like the bonding may try to read out the speed only
at link state changes, so we need to make sure that the link speed is
already set to the expected value when setting the link itself.

Else it may lead to e.g. bond interfaces with wrong link speeds with 10
Mbit/s, despite both links being up with 100Gbit/s.

Fixes: b5fe46c ("tap_manager: set port speed on tap interfaces")


KanjiMonster:
That's not quite it, will need to investigate further.

PR #360, opened 2021-12-16
nl_l3::del_l3_addr: properly resolve vlan interfaces

KanjiMonster:
cnetlink::get_port_id(int) only supports bond and tap devices, but not
vlan interfaces on top of them. for that we need to use
cnetlink::get_port_id(rtnl_link *).

Fixes dropping references for the termination mac flow entries, else
we just pass a port id of 0, which will then fail since the created one
used the actual port id.


jklare:
this targets #361

rubensfig:
This feels like a step in the right direction, however it will not cover all necessary cases.

Following a  routed vlan bridge setup, I considered the following setup.

```
swbridge.10 | swbridge.30
       |                      |
                    swbridge
``` 

Where swbridge.10: 10.0.1.1/24 and swbridge.20 has 10.0.2.1/24 as IP addresses.

The first test is to change the MAC address of the swbridge.10 interfaces, and then delete the IP address. The following output is seen in flowtable 20

```
--  inPort:mask = 0:0x0 etherType = 0x0800 destMac:mask = 0200.0000.0001:ffff.ffff.ffff vlanId:mask = 0x100a:0x1fff (VLAN 10) | GoTo = 30 (Unicast Routing) outPort = 0 (Physical)  | priority = 2 hard_time = 0 idle_time = 0 cookie = 30
```
Before and after the deletion on the IP address. This leads me to think that there is a broken interaction between this code path and https://github.com/bisdn/basebox/blob/5805132c1fbcfccdda67e7f45dd1a3ebe46b6945/src/netlink/nl_l3.cc#L1225

KanjiMonster:
> This feels like a step in the right direction, however it will not cover all necessary cases.
> 
> Following a routed vlan bridge setup, I considered the following setup.
> 
> ```
> swbridge.10 | swbridge.20
>        |               |
>      swbridge
> ```
> 
> Where swbridge.10: 10.0.1.1/24 and swbridge.20 has 10.0.2.1/24 as IP addresses.
> 
> The first test is to change the MAC address of the swbridge.10 interfaces, and then delete the IP address. The following output is seen in flowtable 20
> 
> ```
> --  inPort:mask = 0:0x0 etherType = 0x0800 destMac:mask = 0200.0000.0001:ffff.ffff.ffff vlanId:mask = 0x100a:0x1fff (VLAN 10) | GoTo = 30 (Unicast Routing) outPort = 0 (Physical)  | priority = 2 hard_time = 0 idle_time = 0 cookie = 30
> --  inPort:mask = 0:0x0 etherType = 0x86dd destMac:mask = 0200.0000.0001:ffff.ffff.ffff vlanId:mask = 0x100a:0x1fff (VLAN 10) | GoTo = 30 (Unicast Routing) outPort = 0 (Physical)  | priority = 2 hard_time = 0 idle_time = 0 cookie = 31
> ```
> 
> Before and after the deletion on the IP address. This leads me to think that there is a broken interaction between this code path and
> 
> https://github.com/bisdn/basebox/blob/5805132c1fbcfccdda67e7f45dd1a3ebe46b6945/src/netlink/nl_l3.cc#L1225
> 
> Similarly, flushing the IPv6 link local address does not remove the entry on the table.

This looks like an issue in the code handling mac address changes, not deletion of ip addresses.

Also you really need to provide more output here, but maybe rather create a separate ticket for that?

rubensfig:
Create #362 after bug was discovered. 

Otherwise LGTM

rubensfig (approved):

Issue #362, opened 2022-01-11
Termination MAC entries not being removed after changing SVI MAC address


rubensfig:
## Expected Behavior

Following a  routed vlan bridge setup, I considered the following setup.

Where swbridge.10: 10.0.1.1/24 and MAC address AA:BB:CC:DD:EE:FF.

Flowtable 20 will have the following entry

```
--  inPort:mask = 0:0x0 etherType = 0x0800 destMac:mask = AABB.CCDD.EEFF:ffff.ffff.ffff vlanId:mask = 0x100a:0x1fff (VLAN 10) | GoTo = 30 (Unicast Routing) outPort = 0 (Physical)  | priority = 2 hard_time = 0 idle_time = 0 cookie = 30
--  inPort:mask = 0:0x0 etherType = 0x86dd destMac:mask = AABB.CCDD.EEFF:ffff.ffff.ffff vlanId:mask = 0x100a:0x1fff (VLAN 10) | GoTo = 30 (Unicast Routing) outPort = 0 (Physical)  | priority = 2 hard_time = 0 idle_time = 0 cookie = 31
```

After changing the MAC address to 02:00:00:00:00:01, we see the following entries in the same table

```
--  inPort:mask = 0:0x0 etherType = 0x0800 destMac:mask = 0200.0000.0001:ffff.ffff.ffff vlanId:mask = 0x100a:0x1fff (VLAN 10) | GoTo = 30 (Unicast Routing) outPort = 0 (Physical)  | priority = 2 hard_time = 0 idle_time = 0 cookie = 30
--  inPort:mask = 0:0x0 etherType = 0x86dd destMac:mask = 0200.0000.0001:ffff.ffff.ffff vlanId:mask = 0x100a:0x1fff (VLAN 10) | GoTo = 30 (Unicast Routing) outPort = 0 (Physical)  | priority = 2 hard_time = 0 idle_time = 0 cookie = 31
```

Deleting the IP address should remove the IPv4 entry thus showing 

```
--  inPort:mask = 0:0x0 etherType = 0x86dd destMac:mask = 0200.0000.0001:ffff.ffff.ffff vlanId:mask = 0x100a:0x1fff (VLAN 10) | GoTo = 30 (Unicast Routing) outPort = 0 (Physical)  | priority = 2 hard_time = 0 idle_time = 0 cookie = 31
```

Deleting the Ipv6 link local address should not show any entry wrt to this VLAN

## Actual Behavior

[...]

PR #363, opened 2022-01-20
Fix tmac entries not being removed when cleaning up the IP address

rubensfig:
## Description
Deleting IP addresses from interfaces should remove references from termination MAC table, to ensure correct cleanup of state. 
As described in #362, by changing the MAC address on an interface, and attempting to delete the IP address of this interface, the entries do not seem to get removed. This is due to the cached link reference inside the address object that is about to be deleted. This reference is not updated after the MAC address change, which leads to wrong state being propagated.  

## Motivation and Context

It is necessary for baseboxd to ensure that the state on the switch is cleanup correctly, and no flowtables are left pending.

## How Has This Been Tested?
```
ip address add 10.0.0.1/24 dev port1
ip link set port1 up 
ip link set address 02:01:01:02:01:02 dev port1
ip address delete 10.0.0.1/24 dev port1
```

Refer to #362 for further info on the output to be expected.

### DIFF
--- a/src/netlink/nl_l3.cc
+++ b/src/netlink/nl_l3.cc
@@ -1237,10 +1237,6 @@ int nl_l3::del_l3_termination(uint32_t port_id, uint16_t vid,
   }
 
   // check if not found
-
-  // Modifying the previously existing entries on the VLAN table requires
-  // removing the existing entry, since OFDPA requires that the VLAN must be
-  // configured with the same VRF
   if (i == termination_mac_mapping.end()) {
     LOG(WARNING)
         << __FUNCTION__
--- a/src/netlink/nl_l3.cc
+++ b/src/netlink/nl_l3.cc
@@ -406,7 +406,21 @@ int nl_l3::del_l3_addr(struct rtnl_addr *a) {
 
   int rv = 0;
   int family = rtnl_addr_get_family(a);
-  struct rtnl_link *link = rtnl_addr_get_link(a);
+
+  // The link needs to be obtained from the cache, as the link reference
+  // may be outdated
+  struct rtnl_link *link = nl->get_link(rtnl_addr_get_ifindex(a), AF_UNSPEC);
+
+  // The link object might have already been purged from the cache if the
+  // interface was deleted. Not much we can do here in that case until we start
+  // keeping track of mac addresses ourselves.
+  if (link == nullptr) {
+    LOG(ERROR) << __FUNCTION__
+               << ": unknown link ifindex=" << rtnl_addr_get_ifindex(a)
+               << " of addr=" << OBJ_CAST(a);
+    return -EINVAL;
+  }
+
   uint16_t vid = vlan->get_vid(link);
   uint32_t vrf_id = vlan->get_vrf_id(vid, link);
 


KanjiMonster:
This seems to cause crashes:

```
(gdb) bt
#0  0x00007fe0e7e2de40 in raise () from /lib/libc.so.6
#1  0x00007fe0e7e18528 in abort () from /lib/libc.so.6
#2  0x00007fe0e7e1840f in ?? () from /lib/libc.so.6
#3  0x00007fe0e7e26b32 in __assert_fail () from /lib/libc.so.6
#4  0x000055aec0102672 in basebox::get_link_type (link=link@entry=0x0) at ../git/src/netlink/netlink-utils.cc:74
#5  0x000055aec0122386 in basebox::nl_vlan::get_vid (this=0x55aec21bf930, link=link@entry=0x0) at ../git/src/netlink/nl_vlan.cc:410
#6  0x000055aec0112e89 in basebox::nl_l3::del_l3_addr (this=0x55aec21e1cb0, a=0x7fe0e004ac50)
    at /usr/include/c++/9.3.0/bits/shared_ptr_base.h:1020
#7  0x000055aec00f1d49 in basebox::cnetlink::route_addr_apply (this=0x55aec21c0da0, obj=...) at ../git/src/netlink/nl_obj.h:24
#8  0x000055aec00fc377 in basebox::cnetlink::handle_wakeup (this=0x55aec21c0da0, thread=...) at ../git/src/netlink/cnetlink.cc:778
#9  0x00007fe0e88a25d4 in rofl::cthread::handle_wakeup() () from /usr/lib/librofl_common.so.0
#10 0x00007fe0e88a32b5 in rofl::cthread::run_loop() () from /usr/lib/librofl_common.so.0
#11 0x00007fe0e7fb8ea4 in ?? () from /lib/libpthread.so.0
#12 0x00007fe0e7eebdcf in clone () from /lib/libc.so.6
```

looking at it closer, it looks like the removal of the (probablyIPv6) addrress on the raw swbridge interface is causing this:
```
(gdb) up
#6  0x000055aec0112e89 in basebox::nl_l3::del_l3_addr (this=0x55aec21e1cb0, a=0x7fe0e004ac50)
...
(gdb) p *a->a_link
$3 = {ce_refcnt = 2, ce_ops = 0x7fe0e8a5f500 <link_obj_ops>, ce_cache = 0x0, ce_list = {next = 0x7fe0e00606e8, prev = 0x7fe0e0001fc8}, 
  ce_msgtype = 16, ce_flags = 0, ce_mask = 429696483301, l_name = "swbridge\000\000\000\000\000\000\000", l_family = 7, l_arptype = 1, 
```

src/netlink/nl_l3.cc
```diff
@@ -406,7 +406,18 @@ int nl_l3::del_l3_addr(struct rtnl_addr *a) {
...
+
+  // The link needs to be obtained from the cache, as the link reference
+  // may be outdated
+  struct rtnl_link *link = nl->get_link(rtnl_addr_get_ifindex(a), AF_UNSPEC);
+
+  if (link == nullptr) {
+    LOG(ERROR) << __FUNCTION__
+               << ": unknown link ifindex=" << rtnl_addr_get_ifindex(a)
+               << " of addr=" << OBJ_CAST(a);
+    return rv;
```
KanjiMonster, line 418: rv is zero at this place, you might want to return an error value here

src/netlink/nl_l3.cc
```diff
@@ -406,7 +406,18 @@ int nl_l3::del_l3_addr(struct rtnl_addr *a) {
 
   int rv = 0;
   int family = rtnl_addr_get_family(a);
-  struct rtnl_link *link = rtnl_addr_get_link(a);
+
+  // The link needs to be obtained from the cache, as the link reference
+  // may be outdated
+  struct rtnl_link *link = nl->get_link(rtnl_addr_get_ifindex(a), AF_UNSPEC);
+
+  if (link == nullptr) {
```
KanjiMonster, line 414: A comment here (with a guess why this might happen) would be good. You would assume this couldn't happen, but it obviously does.

src/netlink/nl_l3.cc
```diff
@@ -406,7 +406,19 @@ int nl_l3::del_l3_addr(struct rtnl_addr *a) {
 
   int rv = 0;
   int family = rtnl_addr_get_family(a);
-  struct rtnl_link *link = rtnl_addr_get_link(a);
+
+  // The link needs to be obtained from the cache, as the link reference
+  // may be outdated
+  struct rtnl_link *link = nl->get_link(rtnl_addr_get_ifindex(a), AF_UNSPEC);
+
+  // Ensure the object is obtained correctly
```
KanjiMonster, line 414: I was thinking more of actually explaning why we think this happens, not what this code does (what the code does is obvious). Something like:
```suggestion
  // The link object might have already been purged from the cache if the interface was deleted. Not much we can do here in that case until we start keeping track of mac addresses ourselves.
```

KanjiMonster (approved):
lgtm

PR #366, opened 2022-01-31
nl_bridge: take over fdb entries instead of creating them

KanjiMonster:
A recent fix for the kernel allows fdb entries to be switched to/from
ext_lern, so instead of inspecting every packet on reception, let the
kernel learn and take over fdb entries as ncessary.

## Description

Instead of trying to parse every incoming packet and try to learn the neighbors, just pass them on to the kernel and let it parse them, then handle the new neighbor notification.

This allows dropping the special handling in the RX path for bridges.

## Motivation and Context

Adding support for Broadcom KNET intertfaces means we will not be able to inspect every packet, so we need to instead take over learned neighbors.

## How Has This Been Tested?

Running a pipeline on switches, but should be tested more.

### DIFF
--- a/src/netlink/nl_bridge.cc
+++ b/src/netlink/nl_bridge.cc
@@ -691,7 +691,8 @@ void nl_bridge::add_neigh_to_fdb(rtnl_neigh *neigh, bool update) {
   assert(sw);
   assert(neigh);
 
-  uint32_t port = nl->get_port_id(rtnl_neigh_get_ifindex(neigh));
+  int ifindex = rtnl_neigh_get_ifindex(neigh);
+  uint32_t port = nl->get_port_id(ifindex);
   if (port == 0) {
     VLOG(1) << __FUNCTION__ << ": unknown port for neigh " << OBJ_CAST(neigh);
     return;
@@ -709,8 +710,6 @@ void nl_bridge::add_neigh_to_fdb(rtnl_neigh *neigh, bool update) {
     return;
   }
 
-  bool permanent = true;
-
   // for sure this is master (sw bridged)
   if (rtnl_neigh_get_master(neigh) &&
       !(rtnl_neigh_get_flags(neigh) & NTF_MASTER)) {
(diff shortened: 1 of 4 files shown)


KanjiMonster:
I'm not 100% sure about the vxlan changes (if they are enough), but the test succeeds with it.

src/netlink/nl_bridge.cc
```diff
@@ -730,6 +749,27 @@ void nl_bridge::add_neigh_to_fdb(rtnl_neigh *neigh, bool update) {
             << " vlan=" << (unsigned)vid << ", permanent=" << permanent;
   LOG(INFO) << __FUNCTION__ << ": object: " << OBJ_CAST(neigh);
   sw->l2_addr_add(port, vid, _mac, true, permanent, update);
+
+  if (!permanent) {
+    // take over dynamic entries from kernel
+    rtnl_neigh_set_flags(n.get(), NTF_EXT_LEARNED);
```
rubensfig, line 755: The flag NTF_EXT_LEARNED was introduced in a certain kernel version, should we not catch this somehow?

src/netlink/nl_bridge.cc
```diff
@@ -730,6 +749,27 @@ void nl_bridge::add_neigh_to_fdb(rtnl_neigh *neigh, bool update) {
...
+    assert(msg);
+
+    // send the message and create new fdb entry
+    if (nl->send_nl_msg(msg) < 0) {
+      LOG(ERROR) << __FUNCTION__ << ": failed to send netlink message";
+      return;
+    }
+
+    // cache the entry
+    if (nl_cache_add(l2_cache.get(), OBJ_CAST(n.get())) < 0) {
```
rubensfig, line 768: Bit on confusion on the workflow here.

Wont adding the entry to the cache duplicate the neigh? I assume we already have the neighbor once add_neigh_to_fdb is called

KanjiMonster: Though I don't follow, we are in add_neigh_to_fdb?

src/netlink/nl_bridge.cc
KanjiMonster: NTF_EXT_LEARNED was introduced in 3.19, I think we can assume we are running on a kernel that has it.

src/netlink/nl_bridge.cc
rubensfig: I think I missed the `rtnl_neigh_build_add_request(n.get(), NLM_F_REPLACE, &msg);` above, with the REPLACE flag it makes sense.

PR #367, opened 2022-02-04
prepare support for alternative port device implementations

KanjiMonster:
Abtract out relevant code for managing port devices and replace any references to tap devices to be able to support alternative port implementations in the future.

## Description
To be able to cleanly support e..g knet based port interfaces, we need to refactor our current code to work in a way it does not matter anymore if interfaces are tap interfaces or not.

The steps to reach this are
1. create an interface for creating/destroying/matching port net devices, and sending packets to them.
2. rework basebox::tap_manager to implement that interface.
3. Let all code in baseboxd use the interface instead of tap_manager directly.
4. replace all direct checks for LT_TUN with a check wether the interface has a port_id assigned.

## Motivation and Context
We want to be able to use bcm-knet based interfaces when running on switch, which gives us greatly reduced latencies (on TH+ pings to the switch go from ~3 ms to ~0.2 ms).

This also lays some ground work for eventually supporting other interfaces for configuring switch ASICs than OF-DPA.

## How Has This Been Tested?
Running a pipeline on Agema ag5648s, with no breakage being introduced.

### DIFF
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -880,13 +880,13 @@ void cnetlink::set_tapmanager(std::shared_ptr<port_manager> pm) {
 
 int cnetlink::send_nl_msg(nl_msg *msg) { return nl_send_sync(sock_tx, msg); }
 
-void cnetlink::learn_l2(uint32_t port_id, int fd, basebox::packet *pkt) {
+void cnetlink::learn_l2(uint32_t port_id, basebox::packet *pkt) {
   {
     std::lock_guard<std::mutex> scoped_lock(pi_mutex);
-    packet_in.emplace_back(port_id, fd, pkt);
+    packet_in.emplace_back(port_id, pkt);
   }
 
-  VLOG(2) << __FUNCTION__ << ": got pkt " << pkt << " for fd=" << fd;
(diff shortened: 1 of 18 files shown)


src/netlink/cnetlink.cc
```diff
@@ -1389,36 +1390,35 @@ void cnetlink::link_updated(rtnl_link *old_link, rtnl_link *new_link) noexcept {
...
     break;
   case LT_VRF: // No need to care about the vrf interface itself
   case LT_BRIDGE:
-  case LT_UNSUPPORTED:
     VLOG(2) << __FUNCTION__
             << ": ignoring update (not supported) of old_lt=" << lt_old
             << " old link: " << OBJ_CAST(old_link)
             << ", new link: " << OBJ_CAST(new_link);
     break;
   default:
```
rubensfig, line 1402: Is there no link type for knet interfaces we can add a case for?

KanjiMonster: They are just normal link/ether interfaces, which results in their type(string) being a nullptr. So from a libnl standpoint these are no different than e.g. the management interface.

Also the whole point of this PR is to get rid of special handling for any (port) type, and just let the port_manager handle it.

rubensfig (approved):

Issue #351, opened 2021-11-05
Incorrect ipv6 Termination MAC entries when enslaving port to bridge


hilmarm:
## Expected Behavior
Enslaving a port to a bridge, then tagging a vlan on the port should not create any Termination MAC entries.

```
Table ID 20 (Termination MAC):   Retrieving all entries. Max entries = 512, Current entries = 1.
--  inPort:mask = 7:0xffffffff etherType = 0x86dd destMac:mask = aa39.5aea.5990:ffff.ffff.ffff vlanId:mask = 0x1001:0x1fff (VLAN 1)
| GoTo = 30 (Unicast Routing) outPort = 0 (Physical)  | priority = 2 hard_time = 0 idle_time = 0 cookie = 14

```

## Actual Behavior
Enslaving a port to a bridge, then tagging a vlan on the port does create a Termination MAC entries.

### Set server side port UP before starting switch side configuration.
```
ip link set server_port up
```
### On the Switch run the following commands:

Create vlan_filtering bridge swbridge
```
/sbin/ip link add name swbridge type bridge vlan_filtering 1
```
Set bridge swbridge up
```
/sbin/ip link set swbridge up
```
Attach port_left to the swbridge"
```
/sbin/ip link set {{ port_left }} master swbridge
```
Set port_left attached to swbridge up
```
/sbin/ip link set {{ port_left }} up
```
add vid 10 to port_left
```
/sbin/bridge vlan add vid 10 dev {{ port_left }}
```
Check termination MAC entries
```
client_flowtable_dump 20
```
Gives the output(example given):
```
Table ID 20 (Termination MAC):   Retrieving all entries. Max entries = 512, Current entries = 1.
--  inPort:mask = 7:0xffffffff etherType = 0x86dd destMac:mask = aa39.5aea.5990:ffff.ffff.ffff vlanId:mask = 0x1001:0x1fff (VLAN 1)
| GoTo = 30 (Unicast Routing) outPort = 0 (Physical)  | priority = 2 hard_time = 0 idle_time = 0 cookie = 14
```
## Specifications
  - Version: 1.9.22
  - Platform: agema-ag7648
  - Subsystem:

KanjiMonster:
Does the port have a IP address assigned?

hilmarm:
> Does the port have a IP address assigned?

It does:
```
8: port2: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 state UP qlen 1000
[...]

Issue #364, opened 2022-01-26
ip neighbors are assumed and made routable regardless of actually having a route


KanjiMonster:
## Expected Behavior

Only ip neighbors for which a route exists are being routed (with an appropriate flow).

## Actual Behavior

Any reachable ip neighbor is assumed to be routable and flow tables are setup accordingly (and thus routed in hardware), even if there does not exist a route.

## Steps to Reproduce the Problem

Consider the following setup

```
| Server | port1 <-------------> port1 | Switch |
  lo       10.0.1.2/24     10.0.1.1/24
  10.0.2.1/24
```

```
switch:~$ ip route
10.0.1.0/24 dev port1 proto kernel scope link src 10.0.1.1 
switch:~$ ip route add 10.0.2.0/24 dev port1
switch:~$ ip route
10.0.1.0/24 dev port1 proto kernel scope link src 10.0.1.1 
10.0.2.0/24 dev port1 scope link
switch:~$ ping 10.0.2.1 -I port1
PING 10.0.2.1 (10.0.2.1): 56 data bytes
64 bytes from 10.0.2.1: seq=0 ttl=64 time=1.126 ms
switch:~$ ip neigh show
10.0.2.1 dev port1 lladdr b0:26:28:62:d4:00 REACHABLE
10.0.1.2 dev port1 lladdr b0:26:28:62:d4:00 STALE
switch:~$ client_flowtable_dump 30 | grep 10.
--  etherType = 0x0800 vrf:mask = 0x0000:0x0000 dstIp4 = 10.0.1.0/255.255.255.0 dstIp6 = ::/:: | GoTo = 60 (ACL Policy) outPort = CONTROLLER (Reserved)  | priority = 2 hard_time = 0 idle_time = 0 cookie = 11
--  etherType = 0x0800 vrf:mask = 0x0000:0x0000 dstIp4 = 10.0.1.2/255.255.255.255 dstIp6 = ::/:: | GoTo = 60 (ACL Policy) groupId = 0x20000001 | priority = 3 hard_time = 0 idle_time = 0 cookie = 14
--  etherType = 0x0800 vrf:mask = 0x0000:0x0000 dstIp4 = 10.0.2.0/255.255.255.0 dstIp6 = ::/:: | GoTo = 60 (ACL Policy) outPort = CONTROLLER (Reserved)  | priority = 2 hard_time = 0 idle_time = 0 cookie = 17
[...]

PR #374, opened 2022-03-09
nl_l3: only route l3 neighs if we have a route for them

KanjiMonster:
Make sure we do not add host entries for l3 neighs for which we do not
have a route until we a route gets added, and drop the host entries
again when the route gets removed.

To do this, we need to start keeping track of all l3 neighs we currently
handle. To make this easy, use set, and reuse the nh_stub to store the
neighbor information.

To allow nh_stub to be used in a set, we need to make it comparable, so
implement the < operator for it.

Using this we can then determine the routability of a l3 neigh we add by
asking the kernel for the route to it, then checking if it has a nexthop
that exist on one of our interfaces.

Only if this is true, add a host route for it and add it to the set of
routed l3 neighbours, in all other cases add it to the set of unrouted
l3 neighbours.

Then when adding a route, add all currently unrouted l3 neighbors
matching this route, and conversely, when removing a route remove all
matching l3 neighbors that are currently routed.

Fixes l3 neighbours learned on an additional on-link staying reachable
after removing the on-link route.

## How Has This Been Tested?
Using the commands from https://github.com/bisdn/basebox/issues/364 and verified that no host entries get added without a on-link route, and the host entry appearing and vanashing when adding and removing the on-link route.

Running a pipeline on ag5648 to check for regressions.

### DIFF
--- a/src/netlink/nl_l3.cc
+++ b/src/netlink/nl_l3.cc
@@ -729,6 +729,18 @@ int nl_l3::add_l3_neigh(struct rtnl_neigh *n) {
   }
 
   if (add_host_entry) {
+    // unless we have route, we should not route l3 neighbors
+    struct nh_stub nh {
+      addr, rtnl_neigh_get_ifindex(n)
+    };
+
+    if (is_l3_neigh_routable(n)) {
+      routable_l3_neighs.emplace(nh);
+    } else {
+      unroutable_l3_neighs.emplace(nh);
+      return -ENETUNREACH;
+    }
+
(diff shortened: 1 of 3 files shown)


KanjiMonster:
IPv6 link-local will need some special handling, will fix that tomorrow.

KanjiMonster:
Fixed IPv6 by ignoring IPv6 LL addresses; these should be unroutable anyway.

rubensfig (approved):

rubensfig:
Closes https://github.com/bisdn/basebox/issues/351.

PR #377, opened 2022-04-26
add a post exit script to remove KNET interfaces

KanjiMonster:
Make sure that all KNET interfaces are removed after baseboxd stopped,
so restarting baseboxd (either intentional or because it crashed) does
not fail due to already existing knet interfaces.

This aligns the behaviour with how it works with the default tap
interfaces, which get removed when baseboxd closes.

The tap interfaces are removed automatically since they are owned by
baseboxd and dropped once baseboxd goes away, but this does not work for
the KNET interfaces, since they are neither owned by baseboxd, nor do
they have any facility to remove them automatically.

The proper way to fix this would be for baseboxd handle sigterm and
gracefully remove them, but the code is currently not equipped well for
that, and will require a larger rework.

As an interim smaller solution, add a python script that directly talks
with the KNET kernel module's ioctl interface to remove all interfaces
and filters, with the exception of the default RxAPI filter, which needs
to be kept.

The IOCTL interface definition is taken from

* https://github.com/Broadcom-Network-Switching-Software/OpenBCM/blob/0b149ddfa3878e65eb217a11dddb999d3e205d03/sdk-6.5.24/include/kcom.h
* https://github.com/Broadcom-Network-Switching-Software/OpenBCM/blob/0b149ddfa3878e65eb217a11dddb999d3e205d03/sdk-6.5.24/systems/linux/kernel/modules/include/bcm-knet.h

### DIFF
--- a/scripts/baseboxd-knet-reset.py
+++ b/scripts/baseboxd-knet-reset.py
@@ -1,0 +1,95 @@
+#!/usr/bin/env python3
+#
+# Helper script for resetting KNET to its initial state by removing all
+# added interfaces and filters, intended to be called after stopping
+# baseboxd.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+
+from ctypes import *
+import fcntl
+import os
+
+class bkn_ioctl_t(Structure):
+    _fields_ = [("rc", c_int),
+                ("len", c_int),
+                ("bufsz", c_int),
+                ("reserved", c_int),
+                ("buf", c_uint64)]
+
+class kcom_msg_hdr_t(Structure):
+    _fields_ = [("type", c_uint8),
+                ("opcode", c_uint8),
+                ("seqno", c_uint8),
+                ("status", c_uint8),
+                ("unit", c_uint8),
+                ("reserved", c_uint8),
+                ("id", c_uint16)]
+
+class kcom_msg_list_t(Structure):
+    _fields_ = [("hdr", kcom_msg_hdr_t),
+                ("fcnt", c_uint16),
+                ("id", c_uint16 * 128)]
+
+KCOM_MSG_TYPE_CMD = 1
+
+KCOM_M_NETIF_DESTROY = 12
+KCOM_M_NETIF_LIST = 13
+
+KCOM_M_FILTER_DESTROY = 22
+KCOM_M_FILTER_LIST = 23
+
+# Used for NETIF_DESTROY and FILTER_DESTROY.
+#
+# Both use only the id field of the header to identify the filter/netif
+# to destroy.
+def delete_kcom_object(fd, obj_id, opcode):
+    hdr = kcom_msg_hdr_t()
+    hdr.type = KCOM_MSG_TYPE_CMD
+    hdr.opcode = opcode
+    hdr.unit = 0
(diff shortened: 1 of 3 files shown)


ideaship (changes requested):

ideaship:
I'd suggest running pylint against the script. It's up to you which issues raised by it you want to address, but it might give you some ideas for improvement.

KanjiMonster:
> I'd suggest running pylint against the script. It's up to you which issues raised by it you want to address, but it might give you some ideas for improvement.

Since this is more or less an adaption of c-code, I don't see any of the pylint suggestions as useful, apart from the superfluous  parens after if.

E.g. too few public functions in the Structures? These are c-struct representation, they doesn't have functions. It doesn't make sense to have functions. So this seems to be mostly noise and/or potentially harmful.

ideaship:
As I said, it's up to you. I know it's mostly noise in this case, but still a good habit to use it.

rubensfig (approved):

scripts/baseboxd-knet-reset.py
```diff
@@ -0,0 +1,92 @@
...
+    io = bkn_ioctl_t()
+    io.len = sizeof(obj_list)
+    io.buf = addressof(obj_list)
+
+    fcntl.ioctl(fd, 0, io)
+
+    return obj_list
+
+if os.path.exists("/dev/linux-bcm-knet"):
+    fd = os.open("/dev/linux-bcm-knet", os.O_RDWR | os.O_NONBLOCK)
```
rubensfig, line 81:
```suggestion
    with  os.open("/dev/linux-bcm-knet", os.O_RDWR | os.O_NONBLOCK) as fd:
       ....
```

scripts/baseboxd-knet-reset.py
```diff
@@ -0,0 +1,92 @@
...
+    hdr.type = KCOM_MSG_TYPE_CMD
+    hdr.opcode = opcode
+    hdr.unit = 0
+    hdr.id = obj_id
+
+    io = bkn_ioctl_t()
+    io.len = sizeof(hdr)
+    io.buf = addressof(hdr)
+
+    fcntl.ioctl(fd, 0, io)
```
rubensfig, line 60: Any error/ exception handling for the ioctl call?

scripts/baseboxd-knet-reset.py
ideaship: Actually, if you want to go that way, drop "if os.path.exists("/dev/linux-bcm-knet")". It's not pythonic.

KanjiMonster: So what's the alternative?

KanjiMonster: Then we would need to do a check in the exception handler for it to know whether this is an expected exception or not (if we are running offswitch, then there will be no /dev/linux-bcm-knet, so it failing to open is expected and shouldn't be presented as an error and silently ignored).

Not sure if this is better.

ideaship:
```python
try:
    fd = os.open("/dev/linux-bcm-knet", os.O_RDWR | os.O_NONBLOCK)
except FileNotFoundError:
    pass
```
I'm not sure it's _better_, but it is closer to what Python code is supposed to work like.

KanjiMonster: After thinking about this, I think this one is nice since I can then add a comment before the pass that this is fine and isn't an error (which I just did).

PR #378, opened 2022-05-19
nl_l3: properly handle default routes

KanjiMonster:
RTM_GETROUTE (as done via `nl_route_query`) does not actually give us the route for a destination net, it gives us the route to a destination host, with an implicit /32 resp. /128. This works by accident for all non-default routes (since the first IP is part of the network), but falls apart for IPv4 default routes, since 0.0.0.0/32 is the any address, which aliases to localhost.

To fix this, actually ask the netlink cache for a route with the appropriate destination. As a bonus side effect, the dst of the route
will already have the correct prefix length, and we do not need to modify it anymore.

In addition, fix mask calculation for prefix length 0 on x86 by avoiding undefined behaviour.

### DIFF
--- a/src/netlink/nl_l3.cc
+++ b/src/netlink/nl_l3.cc
@@ -1715,16 +1715,32 @@ int nl_l3::del_l3_ecmp_route(rtnl_route *r,
 }
 
 void nl_l3::nh_reachable_notification(struct nh_params p) noexcept {
-  nl_route_query rq;
-  auto rt = rq.query_route(p.np.addr);
-  auto ad = rtnl_route_get_dst(rt);
+  std::unique_ptr<rtnl_route, decltype(&rtnl_route_put)> filter(
+      rtnl_route_alloc(), rtnl_route_put);
+  std::deque<rtnl_route *> routes;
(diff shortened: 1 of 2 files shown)


src/netlink/nl_l3.cc
```diff
@@ -1715,8 +1715,27 @@ int nl_l3::del_l3_ecmp_route(rtnl_route *r,
...
+      auto dst = rtnl_route_get_dst(r);
+      if (dst == nullptr || nl_addr_get_prefixlen(dst) == 0) {
+        nl_object_get(OBJ_CAST(r));
+        rt = r;
+        break;
+      }
+    }
+  }
+
   auto ad = rtnl_route_get_dst(rt);
```
rubensfig, line 1739: If there is no default route in an interface, or theres something wrong getting the object, wont this crash?

From https://github.com/thom311/libnl/blob/main/lib/route/route_obj.c#L724.

```
struct nl_addr *rtnl_route_get_dst(struct rtnl_route *route)
{
	return route->rt_dst;
```

rubensfig (approved):

PR #385, opened 2022-07-12
Jogo simplify bridge vlan loop

KanjiMonster:
Simplify the bridge vlan loop for improved readability.

## Description
Simplify the bridge vlan loop by rewriting it into a normal for loop, reducing the amount of variables and indentation. Also use a local variable for the vid mask, and make sure it is an unsigned integer to avoid undefined behavior.

Finally rename the outermost k variable to i for completeness.

## Motivation and Context
More readable code == less bugs.


src/netlink/nl_bridge.cc
```diff
@@ -383,14 +383,15 @@ void nl_bridge::update_vlans(rtnl_link *old_link, rtnl_link *new_link) {
          j = find_next_bit(j, vlan_diff)) {
       // vlan added or removed
       int vid = j - 1 + base_bit;
+      uint32_t vid_mask = 1U << (j - 1);
       bool egress_untagged = false;
 
       // check if egress is untagged
-      if (new_br_vlan->untagged_bitmap[k] & 1 << (j - 1)) {
+      if (new_br_vlan->untagged_bitmap[k] & vid_mask) {
```
ideaship: This commit should really be two commits. One that improves readability and one that fixes the potential for undefined behavior.

ideaship (changes requested):
Code looks good, a couple of nitpicks wrt commits.

src/netlink/nl_bridge.cc
KanjiMonster: Split up.

ideaship (approved):

PR #388, opened 2022-07-25
properly ignore AF_INET6 link objects in libnl 3.6+

KanjiMonster:
properly ignore AF_INET6 link objects in libnl 3.6+ that otherwise cause unexpected behavior in baseboxd.

## Description
libnl 3.6.0 introduced listening for IPv6 configuration changes on links. Due to the design of libnl, these are tracked as additional link objects in the cache with the address family AF_INET6, and they only contain basic link information in addition to the IPv6 configuration.

These can easily show up when using rtnl_link_get(ifindex), which is also used internally by rtnl_addr_get_link(). These incomplete link objects then cannot be properly handled by baseboxd, so we need to make sure to get the right objects from the cache (AF_UNSPEC/AF_BRIDGE).

## How Has This Been Tested?

Running a pipeline with this change and libnl 3.7, and having seen no failures related to this anymore.

### DIFF
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -1271,6 +1271,14 @@ void cnetlink::link_created(rtnl_link *link) noexcept {
   assert(link);
   assert(port_man);
 
+  // libnl 3.6+ sends IPv6 configuration updates as incomplete AF_INET6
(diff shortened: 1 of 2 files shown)


rubensfig (approved):
Code LGTM, tested with the libnl 3.7, and no new issues seen.

PR #390, opened 2022-10-05
nl_bridge: default stp state should be forwarding

hilmarm:
## Description
When deleting a VLAN from an stp bridge, set the port state to BR_STATE_FORWARDING

## Motivation and Context
When removing a VLAN from a bridged port, we currently set its STP state to DISABLED, dropping all traffic on it. When the port then gets removed from the bridge, the VLAN will continue being blocked on this port. To ensure this does not happen, use FORWARDING as the default state. The VLAN flow/group memberships will still take care of filtering the traffic while being a bridge member, so this is fine to do.

This follows a change we did in creation of STG groups, where we switched the default from DISABLED to FORWARDING.

## How Has This Been Tested?
This was tested with our latest testing images, including the latest stg fixes in meta-ofdpa
This was run on the agema-ag7648, but will extend to all our supported platforms.

edit: updated context with better explanation

### DIFF
--- a/src/netlink/nl_bridge.cc
+++ b/src/netlink/nl_bridge.cc
@@ -1113,7 +1113,7 @@ int nl_bridge::add_port_vlan_stp_state(uint32_t port_id, uint16_t vid,
 
 int nl_bridge::del_port_vlan_stp_state(uint32_t port_id, uint16_t vid) {
   // default state when not part of the vlan
-  auto g_stp_state = BR_STATE_DISABLED;
+  auto g_stp_state = BR_STATE_FORWARDING;
 
   LOG(INFO) << __FUNCTION__ << ": set state=" << g_stp_state
             << " ifindex= " << port_id << " VLAN =" << vid;


KanjiMonster:
> When a port vlan is removed from an STP bridge, the default behaviour should
be BR_STATE_FORWARDING, not BR_STATE_DISABLED

Please provide an short reasoning why it should be that state. What issue does this fix? Also you are missing a dot at the end of the sentence.

hilmarm:
> > When a port vlan is removed from an STP bridge, the default behaviour should
> > be BR_STATE_FORWARDING, not BR_STATE_DISABLED
> 
> Please provide an short reasoning why it should be that state. What issue does this fix? Also you are missing a dot at the end of the sentence.

Updated commit and added description to PR

KanjiMonster:
> When a port vlan is removed from an STP bridge, the default behaviour should
be BR_STATE_FORWARDING. Currently we set the state to BR_STATE_DISABLED, which
causes traffic to be blocked on the port with the VLANs that were in the STG
group, from which we removed the port.

What you are basically saying here is "If we remove a VLAN from a port, then the VLAN will get blocked on the port, so we need to set FORWARDING instead of DISABLED", but traffic being blocked for VLANs the port isn't a member of is a desired behavior.

You rather want something along

"When removing a VLAN from a bridged port, we currently set its STP state to DISABLED, dropping all traffic on it. When the port then gets removed from the bridge, the VLAN will continue being blocked on this port. To ensure this does not happen, use FORWARDING as the default state. The VLAN flow/group memberships will still take care of filtering the traffic while being a bridge member, so this is fine to do.

This follows a change we did in creation of STG groups, where we switched the default from DISABLED to FORWARDING."

hilmarm:
Commit message and PR description updated with suggested explanation from @KanjiMonster

KanjiMonster (approved):

PR #393, opened 2022-10-14
mark packets as offloaded when using KNET

KanjiMonster:
Linux will flood/forward any packets it sees on a bridge by default. This can lead to packet duplication if the packet was already forwarded/flooded by the ASIC. This can be avoided by marking packets as forwarding being offloaded, so tell KNET to mark any received packets as such.

To allow disabling it in case of unexpected side effects, add a flag for controlling the behavior.

The implementation is kept as generic as possible, in case we ever implement it for TAP interfaces as well.


rubensfig (commented):
LGTM

PR #394, opened 2022-10-21
cnetlink: fix handling of bond mode changes

KanjiMonster:
Restore an accidentally removed case statement causing bond mode changes to be ignored.

Fixes: 7d10d3b8486f ("replace checks for LT_TUN with check for having a port_id")


rubensfig (approved):
LGTM

PR #396, opened 2022-10-27
 rework termination mac handling

KanjiMonster:
Rework termination mac handling by tying termination mac entries to
interfaces:

* create entries on interface creation
* remove them on interface deletion

with the values of
* port_id, vid 0 for the base port/bond interface (specific port, any vlan)
* port_id, vid for vlan interfaces on top of those (specific port, specific vlan)
* port_id 0, vid 0 for the base bridge interface (any port, any vlan)
* port_id 0, vid for vlan interface on top of bridge (any port, specific vlan)

Since these combinations are unique, we can just add/remove the entries
without the need of having refcounts.

Advantages:
* simplication of code
* ip addresses assigned to lo will just work

Disadvantages:
* higher use of termination mac flows, which are limited (0.5k to 1k),
  so only (256 - ports) to (512 - ports) number of vlan/bridge/bond
  interfaces are supported
* (this is a lie though, since both termination mac entries for an
  interface share the same underlying resource)

Open Questions:
* Do other parts need termination mac entries, or the underlying
  resources?


src/netlink/cnetlink.cc
```diff
@@ -663,23 +665,53 @@ int cnetlink::remove_l3_configuration(rtnl_link *link) {
...
 }
 
+int cnetlink::add_termination_mac(rtnl_link *link) {
+  struct nl_addr *addr = rtnl_link_get_addr(link);
+  auto mac = rofl::caddress_ll((uint8_t *)nl_addr_get_binary_addr(addr),
+                               nl_addr_get_len(addr));
+  uint32_t port_id = get_port_id(link);
+  uint16_t vid = 0;
+
+  if (rtnl_link_is_vlan(link))
```
rubensfig, line 682: In the below default code, VLAN check does not seem to be guarded, is that/ this correct?

https://github.com/bisdn/basebox/blob/00e5761724d45f6ab78d90266cc70bd5e7df996c/src/netlink/nl_l3.cc#L210

rubensfig (commented):
Hey! 

Is it possible to split up the commits?
Thank you

src/netlink/cnetlink.cc
KanjiMonster: Yes, l3 egress interfaces require a valid VID, while termination mac entries don't (no VID means any).

KanjiMonster:
> Hey!
> 
> Is it possible to split up the commits? Thank you

To be honest, I'm not sure if I can split this up without breaking things in between. At best I could break out adding the new functions, but the switch from per-IP-address to per-interface termination macs needs to be atomic.

src/netlink/cnetlink.cc
```diff
@@ -663,23 +665,53 @@ int cnetlink::remove_l3_configuration(rtnl_link *link) {
...
+    vid = rtnl_link_vlan_get_id(link);
+
+  swi->l3_termination_add(port_id, vid, mac);
```
rubensfig: Hey! The old code seems to have error checking and logging of the failed flow entry.

Is there a need to remove the extra logging?

KanjiMonster: That was a left-over from this being a PoC implementation originally.

I added some error logging to the functions.

KanjiMonster:
* added logging
* changed return type to void (nothing checks for the return value, and not sure if we could act on that anyway)

rubensfig (approved):

PR #397, opened 2023-01-25
properly handle route lookups when no default route

KanjiMonster:
When there is no route for a host, the kernel may respond with an error, which we treat as fatal and crash. Fix this by accepting the error as normal.

## Description
When there is no default route, the kernel may respond with -ENETUNREACH or -EHOSTUNREACH instead of a route. libnl then translates this error as -NLE_FAILURE, so treat -NLE_FAILURE as no route found and just return a nullptr.

## Motivation and Context
We should not crash when failing lookups because the management interface is (temporarily) down.

## How Has This Been Tested?

Ran a pipeline on as4630-54pe, which succeeded.

### DIFF
--- a/src/netlink/nl_route_query.h
+++ b/src/netlink/nl_route_query.h
@@ -83,7 +83,7 @@ public:
     if (err < 0)
       LOG(FATAL) << __FUNCTION__ << ":%s" << nl_geterror(err);
 
-    if (nl_recvmsgs_default(sock) < 0)
+    if ((err = nl_recvmsgs_default(sock)) < 0)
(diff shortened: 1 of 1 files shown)


KanjiMonster:
> When there is no default route, the kernel may respond with -ENETUNREACH or -EHOSTUNREACH instead of a route. libnl then translates this error as -NLE_FAILURE, so treat -NLE_FAILURE as no route found and just return a nullptr.

This can be seen by using libnl-3-cli's nl-route-get for an unrachable ip address (e.g. when the management interface is down).

rubensfig (approved):
Locally tested, LGTM.

PR #399, opened 2023-02-24
nl_bridge: always set port stp state

KanjiMonster:
The kernel will set ports to disabled while the bridge interface is down to inhibit forwarding regardless wether STP is enabled for the bridge, and we should be doing the same.

So do not ignore port STP state changes when STP is disabled.

Fixes: ee54b7ae49d2 ("nl_bridge: implement tracking global and pervlan stp state;       * adds struct to track g/pv stp states;         * adds setting of stp state in the ASIC")

## Description

see above

## Motivation and Context

Bridges while down should not allow communication between ports at all.

## How Has This Been Tested?

Pinging between hosts via switch, once with the bridge interface down, once up (e.g. vlan-bridge test).

Before pinging succeeds always regardless of admin state of bridge, after only if the bridge is up.

### DIFF
--- a/src/netlink/cnetlink.h
+++ b/src/netlink/cnetlink.h
@@ -82,10 +82,6 @@ public:
   int get_port_id(rtnl_link *l) const;
   int get_port_id(int ifindex) const;
   int get_ifindex_by_port_id(uint32_t port_id) const;
-  int get_bridge_stp_state() {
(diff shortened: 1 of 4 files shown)


rubensfig (approved):
LGTM, build tested locally.

PR #401, opened 2023-03-22
controller: fix ::egress_port_vlan_remove() for bonds

KanjiMonster:
Bond interfaces use a different L2 Interface group type, so we need to check for bond and handle them in controller::egress_port_vlan_remove(), like we do in controller::egress_port_vlan_add().

Fixes: c9f6de2b6f62 ("bond: expand bonding feature in baseboxd")


rubensfig (approved):
LGTM, build tested locally.

PR #413, opened 2023-09-01
fix port state race properly

KanjiMonster:
Make sure that the link state of ports matches the state after the first port event from OF-DPA.

## Description
https://github.com/bisdn/basebox/pull/412 tried to fix the startup race between the port desc reply and the first link state event. While the solution fixed the case where the link state event was first, it broke the case where the first link state comes in after the port desc reply, as it now was ignored as the very first link state change event is sent as `PORT_STATE_NEW`.

To fix this properly, introduce a new (internal) event type `PORT_EVENT_TABLE` to mark the reply, and use this for port creation, and then treat `PORT_STATE_NEW` and `PORT_STATE_MODIFY` as the same, and let them only update state.

## Motivation and Context

Fixes link states changes being lost on start up.

## How Has This Been Tested?

Run a pipeline on as4610.


rubensfig (approved):
LGTM

PR #417, opened 2023-11-08
cnetlink: just ignore bridges with unsupported configuration

KanjiMonster:
Ignore bridges with unsupported configuration instead of treating it as fatal state and crashing.

## Description

Ignore bridges with unsupported configuration instead of treating it as fatal state and crashing. Do this by keeping a list of ignored bridges, and ignoring them for attach/detatchment until the bridges get deleted.

## Motivation and Context

This avoid baseboxd getting in a broken state, as we won't apply any configuration from these bridges.

This avoids getting in an invalid OF-DPA state requiring a reboot to recover, and potentially going into a crash loop.

## How Has This Been Tested?

create a non vlan-aware bridge, see if attaching a port gets ignored
create a vlan aware bridge, see if attaching a port is processed
create a second vlan aware bridge, see if attaching a different port is ignored

### DIFF
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -20,6 +20,7 @@
 #endif
 #include <netlink/route/link.h>
 #include <netlink/route/link/bonding.h>
+#include <netlink/route/link/bridge_info.h>
 #include <netlink/route/link/vlan.h>
 #include <netlink/route/link/vrf.h>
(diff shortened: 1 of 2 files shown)


KanjiMonster:
Pipeline succeeded on as4630 (\#55417).

rubensfig (approved):
LGTM

PR #421, opened 2023-11-13
drop the need for OBJ_CAST() when logging netlink objects

KanjiMonster:
Provide appropriate ``operator<<``s for various `rtnl_*` structs instead of forcing callers to `OBJ_CAST()` them, and drop `OBJ_CAST()` from all users.

## Motivation and Context

For some netlink structs we have a `operator<<`, for others these must be cast to `nl_object *` via `OBJ_CAST()`. This makes this error prone, as one might forget to hast (which then causes just he object's address to be printed, or add a `OBJ_CAST()` when one isn't needed, which may even cause crashes if the cast object isn't a `nl_object` (e.g. `nl_addr`).

So handle this casting via appropriate `operator<<`s, and reduce the possibilities of errors.

## How Has This Been Tested?

Built and run tested with debugging enabled.

### DIFF
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -313,7 +313,7 @@ struct rtnl_link *cnetlink::get_link(int ifindex, int family) const {
 
       if (nl_object_match_filter(obj.get_old_obj(), OBJ_CAST(filter.get()))) {
         _link = LINK_CAST(obj.get_old_obj());
-        VLOG(1) << __FUNCTION__ << ": found deleted link " << OBJ_CAST(_link);
+        VLOG(1) << __FUNCTION__ << ": found deleted link " << _link;
         break;
       }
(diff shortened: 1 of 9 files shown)


KanjiMonster:
This issue already existed before, but was hidden due to the `OBJ_CAST()`.

KanjiMonster:
Pushed fixes for the unearthed issues.

rubensfig:
This MR will need to be rebased after #417, as it created a merge conflict locally for me.

KanjiMonster:
Yepp, whichever you merge first the other one will need to be rebased.

rubensfig:
Ready for rebase then

KanjiMonster:
> Ready for rebase then

Too slow, already rebased. ;P

rubensfig (approved):

PR #423, opened 2024-01-22
nl_bridge: assume bonds may have nullptr bridge vlans as well

KanjiMonster:
We already assume that `rtnl_link_bridge_get_port_vlan()` may return a `nullptr` for regular bridge vlan updates, so we should do the same when calling `nl_bridge::update_vlans()` for bond attachment.

Therefore move the `empty_br_vlan` replacement to after the bond attachment special case, to ensure `old_br_vlan` and `new_br_vlan` are always valid.

Fixes a crash observed where `rtnl_link_bridge_get_port_vlan()` returned a `nullptr` for the parent bond interface, causing
`nl_bridge::br_vlan_equal()` asserts that both arguments are valid to trigger.

Reported-by: Jan Klare <jan.klare@bisdn.de>


rubensfig (approved):
LGTM

PR #424, opened 2024-01-30
of-dpa: reset switch state on connect

KanjiMonster:
To ensure that the baseboxd and switch state are in sync, reset the switch state on connect:

* delete all flows
* delete all groups
* delete all tunnels and their tenents and nexthops
* delete all bond ports
* reset STG groups to default (all vlans in group 1)

This ensures that the we have a valid state if baseboxd was restarted, either voluntarily (due to a restart) or involuntarily (due to a crash).

### DIFF
--- a/pkg/systemd/sysconfig.template
+++ b/pkg/systemd/sysconfig.template
@@ -14,6 +14,9 @@
 #
 # Mark switched packets as offloaded:
 # FLAGS_mark_fwd_offload=true
+#
+# Clear switch configuration on connect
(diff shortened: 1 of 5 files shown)


KanjiMonster:
Thinking about it a bit, I'll add a knob for it so you can disable this if you want/need to, but it should not stop anyone from testing this.

KanjiMonster:
Added a flag to disable the behaviour, and updated `/src/grpc/proto` to point to the updated master of protobuf-basebox.

rubensfig (approved):

PR #425, opened 2024-02-14
ctapdev: create tap devices with carrier off

KanjiMonster:
Since linux 6.0 it is possible to create tap devices with the carrier off instead of needing to set it off after creation [1].

This prevents a race between bring up and setting the carrier off.

[1] https://github.com/torvalds/linux/commit/195624d9c26b64c6856863da30ec578a790feec4


KanjiMonster:
Depends on https://github.com/bisdn/meta-open-network-linux/pull/151

rubensfig (commented):
LGTM.

Build tested.

rubensfig (approved):

PR #428, opened 2024-03-07
nl_bridge: fix removing aged out neighbors from fdb

KanjiMonster:
When fixing moving neighbors between ports we removed the ifindex from the filter neigh to find a neighbor at any port with the mac address.

But since we use the same neigh to construct the fdb delete message, this caused the message to miss the ifindex, and the kernel was rejecting the deletion. This resulting in us not updating out local cache of known layer 2 neighbors, and failing to relearn them.

Fix this by just using the found entry from our local cache, n_lookup.

Fixes: a566d4caa1b0 ("nl_bridge::fdb_timeout(): check ifindex of hit before deletion")

## How Has This Been Tested?

1. Let the bridge learn a neighbor
2. wait for the neighbor to timeout (after 300 seconds)
3. check with bridge fdb show wether the neighbor still exists

expected behavior: neighbor is removed from fdb

actual behavior: neighbor is still present in fdb

### DIFF
--- a/src/netlink/nl_bridge.cc
+++ b/src/netlink/nl_bridge.cc
@@ -837,7 +837,7 @@ int nl_bridge::fdb_timeout(rtnl_link *br_link, uint16_t vid,
                       rtnl_link_get_ifindex(br_link)) {
     // * remove l2 entry from kernel
     nl_msg *msg = nullptr;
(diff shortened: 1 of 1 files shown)


rubensfig (approved):
LGTM

PR #429, opened 2024-04-25
cnetlink: only check bridge for vlan_filtering on creation

KanjiMonster:
We don't need to check for vlan_filtering being enabled every time we attach a port to the bridge, only on the first attachement of a port to the bridge.

We don't handle vlan_filtering state changes, so even if it was disabled later, we wouldn't be able to do anything about it. So let's only check the state once when populating cnetlink::bridge.

Fixes: f93b9b54f161 ("cnetlink: just ignore bridges with unsupported configuration")


rubensfig (approved):
LGTM

Issue #414, opened 2023-09-04
VxLAN packets are encapsulated with an outer VLAN tag with VID 1 on Trident3


KanjiMonster:
## Expected Behavior

There is no outer VLAN tag on VxLAN packets unless explicitly configured.

## Actual Behavior

On Trident3, there is an outer VLAN tag on VxLAN packets.

## Steps to Reproduce the Problem

Create a VxLAN tunnel, try to send packets over it. When looking at them form the other side, there will be a VLAN tag added to them, while non-VxLAN packets are correctly untagged:

```
08:07:48.026027 f8:8e:a1:e0:bb:14 > 90:e2:ba:61:16:50, ethertype 802.1Q (0x8100), length 114: vlan 1, p 0, ethertype IPv4, (tos 0x0, ttl 45, id 50, offset 0, flags [none], proto UDP (17), length 96)
    192.168.0.2.54443 > 192.168.0.1.4789: VXLAN, flags [I] (0x08), vni 50000
f8:8e:a1:e0:83:49 > 90:e2:ba:61:13:a0, ethertype ARP (0x0806), length 60: Ethernet (len 6), IPv4 (len 4), Request who-has 10.0.0.1 tell 10.0.0.2, length 46
08:07:49.060756 f8:8e:a1:e0:bb:14 > 90:e2:ba:61:16:50, ethertype 802.1Q (0x8100), length 114: vlan 1, p 0, ethertype IPv4, (tos 0x0, ttl 45, id 51, offset 0, flags [none], proto UDP (17), length 96)
    192.168.0.2.54443 > 192.168.0.1.4789: VXLAN, flags [I] (0x08), vni 50000
f8:8e:a1:e0:83:49 > 90:e2:ba:61:13:a0, ethertype ARP (0x0806), length 60: Ethernet (len 6), IPv4 (len 4), Request who-has 10.0.0.1 tell 10.0.0.2, length 46
08:07:50.094093 f8:8e:a1:e0:bb:14 > 90:e2:ba:61:16:50, ethertype 802.1Q (0x8100), length 114: vlan 1, p 0, ethertype IPv4, (tos 0x0, ttl 45, id 52, offset 0, flags [none], proto UDP (17), length 96)
    192.168.0.2.54443 > 192.168.0.1.4789: VXLAN, flags [I] (0x08), vni 50000
[...]

PR #433, opened 2024-05-03
allow configuring the default untagged VLAN for ports

KanjiMonster:
Using unbridged, untagged ports at the same time as VID 1 as default PVID on a bridge may cause unexpected leakage and forwarding of traffic.

Moving to the reserved VID 4095 would avoid this, but we cannot unconditionally use it until we solved #414, as otherwise we would leak packets with VID 4095.

So for now allow changing the VID used internally for ports, but keep the default at 1.

### DIFF
--- a/pkg/systemd/sysconfig.template
+++ b/pkg/systemd/sysconfig.template
@@ -17,6 +17,9 @@
 #
 # Clear switch configuration on connect
 # FLAGS_clear_switch_configuration=true
+#
+# Vlan ID used for untagged traffic on unbridged ports (1-4095):
+# FLAGS_port_untagged_vid=1
 
 ### glog
 #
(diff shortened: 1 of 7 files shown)


pkg/systemd/sysconfig.template
```diff
@@ -17,6 +17,9 @@
 #
 # Clear switch configuration on connect
 # FLAGS_clear_switch_configuration=true
+#
+# Vlan ID used for untagged traffic on unbridged ports (1-4095):
```
rubensfig, line 21: Would it make sense to raise a warning here that this is only used internally? 

I am concerned with the possibility of misconfiguration and people mistaking this field with the systemd-networkd configuration files.

rubensfig:
Tested the patch, and seems to be working as expected. 

Some concerns, however. Like mentioned in the other comment, this significantly raises (IMO) the bar for a misconfiguration between this parameter and systemd-network.

Maybe its more implementation work, but don't we have the state of the swbridge in the cache/ can't we query Linux for the VLAN untagged state?

rubensfig (approved):

rubensfig:
LGTM, all discussions resolved!

PR #434, opened 2024-05-06
examples: pass basic l2 configuration

rubensfig:
https://github.com/bisdn/bisdn-linux/issues/101 has raised the attention to the need of having a set of scripts to setup the switch with a default behaviour that remains.

This PR adds a set of files to enable this behaviour.


examples/networkd/20-bridge.l2-learn.netdev
```diff
@@ -1,5 +1,5 @@
 [NetDev]
```
KanjiMonster, line 1: The installer expects the files in a subdirectory appropriately named, so I would suggest moving them to `examples/networkd/<examplename>/`, where `<examplename>` might be something like `simple-layer2-bridge`.

Also maybe follow the example filenames and numbering from https://docs.bisdn.de/network_configuration/vlan_bridging.html#systemd-networkd.

examples/networkd/simple-l2-bridge/10-swbridge.l2-learn.netdev
```diff
@@ -1,5 +1,5 @@
 [NetDev]
```
KanjiMonster, line 1: Can we drop the `.l2-learn` from the filenames? I think it is unexpected and would only confuse.

KanjiMonster (approved):
LGTM now

PR #436, opened 2024-05-29
fix warnings and enable -Werror

KanjiMonster:
Warnings are important and should be addressed, so make sure they cannot be ignored by enabling -Werror after fixing all warnings.

No functional changes.


rubensfig (approved):
LGTM

PR #438, opened 2024-05-29
adapt mdb handling to fixed upstream working

KanjiMonster:
With a recent fix, libnl will now send the updated object instead of the update as new for callback_v2(). Unfortunately the mdb code relied on getting incremental updates as new, so this change broke mdb handling.

To fix this, rework the mdb handling the following:

1. convert all entries of old and new into separate, ordered sets
2. for any entry in the old, but not new leave the group
3. for any entry in the new, but not old join the group

This with empty sets for old respective new this can also be used new and delete of the object.

### DIFF
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -1796,20 +1796,35 @@ std::deque<rtnl_neigh *> cnetlink::search_fdb(uint16_t vid, nl_addr *lladdr) {
 
 void cnetlink::route_mdb_apply(const nl_obj &obj) {
 
-  switch (obj.get_msg_type()) {
-  case RTM_NEWMDB:
+  switch (obj.get_action()) {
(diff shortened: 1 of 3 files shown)


KanjiMonster:
Added a few more comments and some debug output.

KanjiMonster:
Depends on https://github.com/bisdn/meta-switch/pull/206

KanjiMonster:
This may also as a side effect fix issues when (un-)subscribing to different ip addresses that has the same multicast ll mac.

E.g. with the previous implementation:

1. subscribe to Ip A => ll mac X (join)
2. subscribe to Ip B => ll mac X (join ... again)
3. unsubscribe from Ip A => ll mac X (leave)

would end up with leaving the mc group X despite still being interested in it.

rubensfig (approved):
LGTM, build and run-tested.

PR #439, opened 2024-05-31
cnetlink::get_neighbor(): check pending notification for deleted neighs

KanjiMonster:
When a route and its nexthop (neigh) get removed at the same time, we may process the route removal before the neigh removal.

On route removal we try to unregister for nexthop notifications, and remove any pending l3 interfaces.

Since cnetlink::get_neighbour() only asks the cache, it only returns live neighs, but since the neigh was already removed from the cache, it wont find any.

At the same time, since we did not process the neigh removal yet, we are still registered for nh unreachable notifications, but since from the perspective of the nl_l3::del_l3_route(), the nexthop is already unreachable, so we fail to remove the unreachable notification.

Fix this by also going through the pending neigh deletions to find the reference in case there is a neigh delete pending.

Fixes: dfe44856fe31 ("add netlink functions")

### DIFF
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -649,8 +649,39 @@ struct rtnl_neigh *cnetlink::get_neighbour(int ifindex,
                                            struct nl_addr *a) const {
   assert(ifindex);
   assert(a);
+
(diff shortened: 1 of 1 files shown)


rubensfig (approved):
LGTM

PR #441, opened 2024-06-04
nl_l3: fix nnhs handling a bit

KanjiMonster:
Fix a few issues around number of nexthops handling:

- we rejected/ignored routes with 0 nexthops (`nnhs` = 0) on add, but don't do so in removal (though I'm not convinced this is an actual thing that can happen)
- later on we assume `nnhs` = 0 means no reachable(?) nexthop yet, but we already rejected routes with `nnhs` = 0 earlier, so this can't happen! Instead, check if `l3_interfaces` is empty
- when removing a route, don't use the amount of `l3_interfaces` to determine if this is a ECMP route, but use `nnhs` instead

### DIFF
--- a/src/netlink/nl_l3.cc
+++ b/src/netlink/nl_l3.cc
@@ -2024,7 +2024,7 @@ int nl_l3::del_l3_unicast_route(rtnl_route *r, bool keep_route) {
     return rv;
   }
 
-  if (neighs.size() > 1) {
+  if (nnhs > 1) {
     // get all l3 interfaces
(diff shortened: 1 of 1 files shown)


KanjiMonster:
Note that nnhs just means number of nexthop objects, but a nexthop object may just reference a device for directly connected routes, and not necessarily gateways with IPs.

Technically it would be possible to have ECMP routes with multiple on-link device nexthops, but 0 gateways.

rubensfig (approved):
LGTM

PR #442, opened 2024-06-04
cnetlink: clone objects to ensure they are unchanged

KanjiMonster:
Libnl ops that support updating objects may modify the object, even if we haven't handled the notification yet, causing older stored notifications to use an object "from the future".

Avoid this by cloning the new object. Both nl_object_clone() and nl_object_put() are NULL-safe, so we don't need to do any nullptr checks.


KanjiMonster:
Depends on https://github.com/bisdn/meta-switch/pull/215

rubensfig (approved):
Reviewed and tested

rubensfig (approved):

PR #443, opened 2024-06-07
nl_l3: keep IPv6LL route enabled if any interface still has it

KanjiMonster:
Currently we unconditionally delete the IPv6LL route when it get's deleted from the routing table. But this is a per-interface route, so just because one route is removed does not mean that there are none left.

So the following sequence

    $ ip link set up port1
    $ ip link set up port2
    $ ip link set down port1

will result in the IPv6LL route being disabled, although port1 still has one.

So add a check on deletion if there are any remaining interfaces left with the IPv6LL route, and only actually delete it if there aren't any.

### DIFF
--- a/src/netlink/nl_l3.cc
+++ b/src/netlink/nl_l3.cc
@@ -1968,6 +1968,52 @@ int nl_l3::del_l3_unicast_route(rtnl_route *r, bool keep_route) {
   }
 
   if (!keep_route) {
+    // OpenFlow / OF-DPA does not support per-interface routes, so we can only
+    // have one IPv6LL route. Therefore make sure we only delete it if there are
+    // no interfaces left with one.
+    if (is_ipv6_link_local_address(dst)) {
+      std::unique_ptr<rtnl_route, decltype(&rtnl_route_put)> filter(
+          rtnl_route_alloc(), rtnl_route_put);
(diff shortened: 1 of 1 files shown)


KanjiMonster:
This is more a general issue than just IPv6(LL) and affects any duplicate routes, but with the default IPv6LL routes on each enabled interface this is where it crops up by default.

You have the same issue when assigning IP addresses from the same subnet to multiple interfaces, but a proper fix for that will require libnl changes.

src/netlink/nl_l3.cc
```diff
@@ -1968,6 +1968,50 @@ int nl_l3::del_l3_unicast_route(rtnl_route *r, bool keep_route) {
...
+            nl->is_switch_interface(rtnl_route_nh_get_ifindex(nh))) {
+          keep_route = true;
+        }
+
+        rtnl_route_put(route);
+      }
+    }
+
+    if (keep_route) {
+      VLOG(2) << __FUNCTION__ << ": IPv6LL still has interfaces";
```
rubensfig, line 2011: Can we reword the message a bit?

KanjiMonster: Reworded it a bit.

rubensfig (approved):
LGTM

PR #444, opened 2024-06-21
add support for changing the bridge ageing time

KanjiMonster:
Add support for changing bridge ageing time.

Depends on libnl 3.10.

Compile tested only.


KanjiMonster:
Doesn't work, as the kernel does not send updates about changes in ageing time *sigh*

PR #446, opened 2024-08-09
nl_bridge: flush our fdb entries on vlan removal

KanjiMonster:
The kernel refuses to remove fdb entries for VLANs that are not configured, but at the same time won't remove permanent entries on VLAN deletion, including extern_learn entries.

Fortunately recent kernels gained the ability for bulk deletion, so add a new helper to send a flush of all extern_learn entries from a deleted vlan from a certain port.

Using this, we can make sure no entries are left behind in the removed vlan.

Any removed entry will trigger a DEL_NEIGH message from the kernel, but since we already removed the neigh from the cache and removed the flow, we do not need to handle it again here, so add a check before attempting to handle the deleted neigh.

This was tested via:

on switch:

```
$ ip link add swbridge type bridge vlan_filtering 1 vlan_default_pvid 0 $ ip link set dev port5 master swbridge
$ ip link set dev port5 up
$ bridge vlan add dev port5 vid 2 pvid untagged
```

(connect something to port5, let it generate packets)

```
$ bridge fdb show dev port5 | grep extern_learn
0c:c4:7a:9c:29:fc vlan 2 extern_learn master swbridge
$ bridge vlan del dev port5 vid 2
```

Before:

```
$ bridge fdb show dev port5 | grep extern_learn
0c:c4:7a:9c:29:fc vlan 2 extern_learn master swbridge
```

(entry still present)

After:

```
$ bridge fdb show dev port5 | grep extern_learn
```

(entry gone)

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -29,6 +29,7 @@ sources = files('''
   src/netlink/nl_hashing.h
   src/netlink/nl_interface.cc
   src/netlink/nl_interface.h
+  src/netlink/nl_fdb_flush.h
   src/netlink/nl_l3.cc
   src/netlink/nl_l3.h
   src/netlink/nl_l3_interfaces.h
--- a/src/netlink/nl_bridge.cc
+++ b/src/netlink/nl_bridge.cc
@@ -25,6 +25,7 @@
 #include "cnetlink.h"
 #include "netlink-utils.h"
 #include "nl_bridge.h"
+#include "nl_fdb_flush.h"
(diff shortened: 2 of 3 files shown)


KanjiMonster:
For us being able to handle the DEL_NEIGH messages, we would need to do reference counting for the vlan group for bridges as well - currently we (try to) delete it directly on vlan removal.

rubensfig (approved):

PR #447, opened 2024-08-12
meson.build: properly get systemd unit path

KanjiMonster:
Instead of hardcoding the systemd unit path, extract it via a systemd dependency.


rubensfig (approved):
LGTM

PR #449, opened 2024-08-20
nl_l3::del_l3_neigh_egress(): remove created VLAN as well

KanjiMonster:
In add_l3_neigh_egress() we add a VLAN (reference), but we never remove it in del_l3_neigh_egress(). This creates a reference imbalance, and causes the vlan to be never deleted once there was a neighbor on an interface.

Fix this by removing the VLAN on neigh removal as well.


KanjiMonster:
Pipeline 76747 with changes applied ran successfully.

rubensfig (approved):

PR #450, opened 2024-08-23
set rx rate to unlimited for knet

KanjiMonster:
Since KNET uses a different channel for packets than OpenFlow and is much more efficient, higher packet rates than 1024 (the default) is easily doable.

So set the rx rate to unlimited by default when using KNET, but leave it at 1024 pps for TAP.

In case this causes issue, add a flag to set a custom limit.


pkg/systemd/sysconfig.template
```diff
@@ -20,6 +20,12 @@
 #
 # Vlan ID used for untagged traffic on unbridged ports (1-4095):
 # FLAGS_port_untagged_vid=1
+#
+# Set a Packets per Seconds rate limit for traffic to controller.
+#  -1 = auto (unlimited for KNET, 1024 for TAP)
+#   0 = force unlimited
+# Default is auto.
+# FLAGS_rx_rate_limit=-1
```
jklare, line 28: How do I set a limit of 1234 ? FLAGS_rx_rate_limit=1234 ?

jklare (approved):

PR #456, opened 2025-01-08
1.x: nl_l3::add_l3_neigh_egress(): handle vrf_id as nullptr

KanjiMonster:
nl_l3::add_l3_neigh_egress()'s signature sets nullptr as the default value for vrf_id, but later accesses it unconditionally.

Fix this by only setting it if set.

Fixes: 7788e2d82804 ("feat: enable IP addressses on Bridge Virtual Interfaces; enable multiple creation of VRF interfaces and attachement of BVI to these interfaces; example script for configuration of the VRFs")
Fixes: 9db220e5c0bf ("nl_l3: use add_l3_neigh_egress() instead of open coding it")

(cherry picked from commit 0157c07e1f875805c05da404b927170d274a37c3)


rubensfig (approved):
LGTM

PR #457, opened 2025-01-17
add support for locked bridge ports and locked fdb entries

KanjiMonster:
Add support for the `LOCKED` flag of switch ports and fdb entries:

* `LOCKED` ports do not learn and do not forward packets from unknown mac addresses, so disable learning and forwarding for them.
* `LOCKED` fdb entries are not used for forwarding, but mark a mac address as "seen" on a port. Since this is basically like they weren't learnt, ignore these entries and only add them to OF-DPA once they lose their `LOCKED` flag.

To be able to do so, we need to enhance OF-DPA with the ability to do fine grained learning configuration. Since OF-DPA is old, model the internal interface after SAI [1], with a two modifications:

* add a mode that corresponds to what we currently use, where we log and forward
* add a method for configuring learning behavior for source port violations (i.e. a neighbor moved to a different port)

[1] https://github.com/opencomputeproject/SAI/blob/5ff3424512031258ed2431b06b64beea5bc36712/inc/saibridge.h#L39

### DIFF
--- a/src/netlink/cnetlink.cc
+++ b/src/netlink/cnetlink.cc
@@ -1750,8 +1750,24 @@ void cnetlink::neigh_ll_created(rtnl_neigh *neigh) noexcept {
 
 void cnetlink::neigh_ll_updated(rtnl_neigh *old_neigh,
                                 rtnl_neigh *new_neigh) noexcept {
-  if (!check_ll_neigh(old_neigh) || !check_ll_neigh(new_neigh))
+  bool handle_old = check_ll_neigh(old_neigh);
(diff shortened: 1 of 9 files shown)


jklare:
Is it wise to use the term "sai" within our code here, while actually just miming certain behaviours of it and not using SAI in any way, shape or form? I would think that this might be pretty confusing for somebody stumbling over this function. Maybe it would be better to only use "sai" in a comment?

jklare:
nvm, we seem to have used this before, so this is not the point to discuss it

jklare (approved):

PR #462, opened 2025-02-18
fix and extend the bumpversion* make targets

KanjiMonster:
Fix and extend the make targets for bumping the version:

1. fix major version bumps to reset the minor version to 0
2. always include the patchver, even when 0 to match existing release versions
3. add a new patchversion bump make target (e.g. 2.3.1 -> 2.3.2)
4. ensure a version bump commit only includes a VERSION file change
5. also sign the version bump commit (which we enforce for the tag as well)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,7 @@ bumpversionminor bumpversionmajor:
 	@IFS='.' read -r -a ver <<< "$(VERSION)"; \
 	if [ "$(BUMP)" = "major" ]; then \
 	  ver[0]=$$(($${ver[0]} + 1)); \
+	  ver[1]=0; \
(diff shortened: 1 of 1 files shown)


KanjiMonster:
Since each version bump is supposed to be backed by a tag as well, we may also consider creating tag automatically and not requiring an extra make tag call.

ideaship (commented):
So far, we have released major releases `1.0.0` and `2.0.0`. Are you sure the next major release should be `3.0`?

KanjiMonster:
> So far, we have released major releases `1.0.0` and `2.0.0`. Are you sure the next major release should be `3.0`?

Good point, updated the versioning to match what we actually did the last few years.

ideaship (approved):
LGTM

PR #464, opened 2025-02-20
1.x: fix and extend the bumpversion* make targets

KanjiMonster:
Backport of https://github.com/bisdn/basebox/pull/462.

While we won't need to create new major releases from this branch, we do want minor and especially patch releases, so let's add these here as well.


ideaship (approved):
LGTM

PR #465, opened 2025-02-20
meson.build: fix finding libs

KanjiMonster:
meson complains that when searching for libs, you should omit the libprefix:

```
Run-time dependency libgflags found: NO (tried pkgconfig and cmake)
WARNING: find_library('libgflags') starting in "lib" only works by accident and is not portable
Library libgflags found: YES
```

So fix this by dropping the lib prefix from any cppc.find_library() calls.

The fact that we even landed here also showed that the initial dependency() name was wrong, so fix both glog and gflags, which do not have a lib prefix for their dependency names.

This is not true for libnl and libsystemd though, they still need the lib prefix for looking them up as a dependency.

After:

```
Run-time dependency gflags found: YES 2.2.2
...
Run-time dependency glog found: YES 0.7.1
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -98,28 +98,28 @@ grpc_cpp = find_program('grpc_cpp_plugin')
 librofl_common = dependency('rofl_common')
 librofl_ofdpa = dependency('rofl_ofdpa')
 
-libgflags = dependency('libgflags', required: false)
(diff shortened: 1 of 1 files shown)


ideaship (approved):
LGTM

PR #470, opened 2025-03-19
nl_l3: work around kernel sending routes using nhid with no nexthops

KanjiMonster:
In rare instances it can happen that a route update targeting a nexthop object in kernel gets send to userspace without any nexthop information set exept the nexthop id.

Since we currently rely on routes having this information, this breaks us applying the route, since we do not know what the actual nexthop is.

Fix this by tracking nexthop objects and using the cache as a fallback for looking up the nextop data of routes by the nexthop id.


KanjiMonster:
Pipeline with the code applied ran successful on ag7648 (87883)

jklare (approved):

PR #487, opened 2026-03-04
nl_bridge: do not try to flush fdb on removed interfaces

KanjiMonster:
When removing a VLAN from a bridge port, we flush the fdb entries from the kernel. When removing a port form the bridge, we also remove all VLANs from that port.

When we try to flush the fdb for that port, the kernel refuses to do so since the port is not a bridge member anymore, and returns an error, causing baseboxd to treat this as fatal.

Since removing a port from a bridge will automatically delete all fdb entries for all vlans, there is no need for us to manually delete them, so skip flushing the fdb if we are in the interface removal path, i.e. new_link is a nullptr.


rubensfig (changes requested):

rubensfig (approved):
Tested on AG7648, LGTM.

PR #488, opened 2026-05-08
increase OpenFlow connection timeouts and make them configurable

KanjiMonster:
OpenFlow says:
    
> "OpenFlow messages are processed out of order (see 6.2), and the processing of some requests by the switch may take a long time, therefore a controller must never terminate a connection because a request is taking too much time. An exception to that rule is for echo replies, a controller or a switch may terminate a connection if the reply for an echo request it sent takes too much time, however there must be a way to disable that feature and the timeout should be large enough to accommodate a wide variety of conditions."
    
A maximum of 4 seconds (1 lifetime + 3 echo) is not much if the OpenFlow connection is conguested due to a large amount of messages to process, so double the echo timout.
    
Since rofl-common also sends out echo requests on long-standing barrier requests, but barrier-requests inhibit processing messages until completed, also increase the lifetime check significantly to allow barrier_requests to take up to 16 seconds, not 4.
    
This should significantly increase the stability of the OpenFlow connection under stress conditions.

In case the chosen limits are insufficient or cause issues otherwise, make the values configurable.

### DIFF
--- a/pkg/systemd/sysconfig.template
+++ b/pkg/systemd/sysconfig.template
@@ -26,6 +26,12 @@
 #   0 = force unlimited
 # Default is auto.
 # FLAGS_rx_rate_limit=-1
+#
+# Set OpenFlow echo request timeout in seconds:
+# FLAGS_of_timeout_echo=6
+#
+# Set OpenFlow idle delay for sending echo requests:
+# FLAGS_of_timeout_lifecheck=10
 
 ### glog logging configuration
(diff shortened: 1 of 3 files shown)


KanjiMonster:
Dropped the reference to the not existing `of_txqueue_max_size` variable.

b3n-l (approved):

rubensfig (approved):

PR #491, opened 2026-09-17
cnetlink: delay VXLAN registration to bridge attachment

KanjiMonster:
We do not support VXLAN links outside of bridges, so it does not make sens to register them before being attached to a bridge.

Therefore delay handling VXLAN links until they get attached to a bridge under our control.

This ensures we only try to register VXLAN links after we are connected via OpenFlow, never before.


hwoesner (approved):
this looks straightforward and should not break anything existing.

